PR #2, opened 2016-03-15
Add "Print student reports" button

pjanik:
We need to render a different report type behind the scenes - per student instead of per class. This PR implements that, add a new button and bunch of `@media print` styles. It also includes a small refactoring of data structure and components. 

Deployed demo:
http://concord-consortium.github.io/portal-report/


knowuh:
Took me a while to get used to the ES6.  But this looks good.

PR #3, opened 2016-03-15
Change 'responses' to 'answers'

pjanik:
This PR includes https://github.com/concord-consortium/portal-report/pull/2, so the #2 should be merged first (I did it, so the new code is updated too).


knowuh:
@pjanik  #2 is merged now, if you want to rebase this.

knowuh:
on the other hand, I just reviewed commit 86f5c37 (which I think was the only commit not in #2 ) and it looked good.  :+1:

PR #7, opened 2016-03-16
fix: Names are not still hidden when you re-open the report

knowuh:
small PR when you get a chance @pjanik 

This small fix obfuscates student names when first loaded if `anonymousReport` is true.  

[#115799893]

https://www.pivotaltracker.com/story/show/115799893

### DIFF
--- a/js/core/report-data.js
+++ b/js/core/report-data.js
@@ -43,9 +43,7 @@ export function question(state, key) {
   const question = state.get('questions').get(key.toString())
                   // Sort answers by student name, so views don't have to care about it.
   return question.set('answers', question.get('answers').map(key => answer(state, key))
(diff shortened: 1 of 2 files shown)


pjanik:
@knowuh, good catch, I broke it recently when I was restructuring reducers / data.
I think it would be good to either extract anonymization to a separate function or chain actions - e.g. `RECEIVE_DATA` could automatically trigger `SET_ANONYMOUS`. I've implemented the former as it's simpler and also I've done the same thing for the question filtering (as new data effectively sets questions filter, so I guess it's good to be consistent). I feel it would be nice to split reducers into smaller files soon, especially if we add more actions and helper functions.

knowuh:
Looks good. :+1:

Issue #4, opened 2016-03-16
Inefficient component updates


pjanik:
Currently, the report data is regenerated each time we change the app state using this function:
https://github.com/concord-consortium/portal-report/blob/master/js/core/report-data.js
This is causing that each React component is updated on every, tiny change of the report state (e.g. answer selected). React's Virtual DOM detects that most of the DOM tree hasn't changed and the browser won't re-render everything though, so in practice it still works quite fast. However it may affect performance at some point, especially when we have a lot of components (=> a lot of students and answers).

Other developers have had this problem already:
https://github.com/reactjs/redux/issues/815
It looks like I've implemented approach 2. from the first post, while it would have been better to implement 1. We would need connected components (containers) wrapping dumb components and fetching data for them from the flat app state.
Here's a simple example of such approach:
https://github.com/reactjs/redux/tree/master/examples/tree-view

PR #9, opened 2016-03-24
If nothing is selected, the report should always show everything.

knowuh:
@pjanik   Please have a look, and let me know if I abused the reducer pattern.  

It seems there are some operations which we will almost always want to apply to our state, when computing display values for example. I had thought maybe to do something more like this:

```
export default function reducer(state = Map(), action) {
  let report = report(state.get('report'), action)
  report = setHasSelectedQuestions(report)
  return Map({
    data: data(state.get('data'), action),
    report: report
  })
}
```

But instead decided to follow the patterns closer to what was being done for `setVisibilityFilterActive` 

[#116249337]

https://www.pivotaltracker.com/story/show/116249337

### DIFF
--- a/js/calculations/index.js
+++ b/js/calculations/index.js
@@ -1,0 +1,6 @@
+
+// return true if no questions are checked for compare view
+export function noSelection(reportState) {
+  const selectCount = reportState.get('questions').filter((question,id) => question.get('selected') === true)
+  return selectCount.size === 0
+}
--- a/js/components/answer.js
+++ b/js/components/answer.js
@@ -18,6 +18,11 @@ const AnswerComponent = {
 export default class Answer extends Component {
   render() {
     const { answer, alwaysOpen } = this.props
+    if (typeof answer === 'undefined') {
+      // TODO:  This should be set, but in the case of sequences
+      // it seems its not. TBD later.
+      return <div>No response</div>
+    }
     const AComponent = AnswerComponent[answer.get('type')]
     if (!AComponent) {
       return <div>Answer type not supported.</div>
--- a/js/components/question-for-student.js
+++ b/js/components/question-for-student.js
@@ -16,6 +16,13 @@ export default class QuestionForStudent extends Component {
     onSelectChange(question.get('key'), event.target.checked)
   }
 
+
+  get prompt() {
+    // see discussion here: https://facebook.github.io/react/tips/dangerously-set-inner-html.html
+    const prompt = this.props.question.get('prompt') || this.props.question.get('name')
+    return {__html: prompt}
+  }
+
   render() {
     const { question, student, number } = this.props
     const studentId = student.get('id')
@@ -26,7 +33,7 @@ export default class QuestionForStudent extends Component {
           <input type='checkbox' checked={question.get('selected')} onChange={this.handleCheckboxChange}/>
           Question #{number}
         </div>
-        <div className='prompt'>{question.get('prompt') || question.get('name')}</div>
+        <div className='prompt' dangerouslySetInnerHTML={this.prompt}/>
         <Answer answer={answer}/>
       </div>
     )
--- a/js/components/report.js
+++ b/js/components/report.js
@@ -70,7 +70,7 @@ export default class Report extends Component {
   render() {
     const { report, showSelectedQuestions, showAllQuestions, setAnonymous } = this.props
     const isAnonymous = report.get('anonymous')
-    const showSelectedDisabled = !report.get('has_selected_questions')
+    const showSelectedDisabled = report.get('noSelection')
     return (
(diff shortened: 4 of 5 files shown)


pjanik:
I'd say that `has_selected_questions` is a redundant variable, as it can be calculated from other state (questions). It might make sense to cache things like this sometimes, but I don't think that performance would be an issue here (we're dealing with a small number of questions).
You may notice that question visibility is also redundant and that bothers me a bit. However, it just simplifies "dumb" components that don't have to use any logic to determine whether question is visible or not, and they don't need access to the filter state. However, if we fixed https://github.com/concord-consortium/portal-report/issues/4, I'd rather remove this calculation from the reducer and all and leave it for the "smart" container that could have an easy access to the whole state.

So, instead of:
```javascript 
const noneSelected = !state.get('has_selected_questions')
```
in `setVisibilityFilterActive`, I'd rather calculate it dynamically:
```javascript 
const noneSelected = Object.keys(questions).filter(key => questions[key].selected === true).length === 0
```
BTW, I guess this statement could be simplified a bit too if you used ImmutableJS structure (so, if you didn't call `.toJS()`):
```javascript
const noneSelected = state.get('questions').filter((id, question) => question.selected === true).length === 0
```
In general I would avoid calling `.toJS()` too often. ImmutableJS structures mimic equivalent mutable ES6 structures (List, Map, [Set](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Set), etc.), so the API is pretty much the same (e.g. filtering, forEach and so on).

knowuh:
Piotr,  Thanks for the thoughtful feedback.  I took all your comments about using immutable collections methods, and ES6 tips.  I tried also putting the logic closer to the views, but in this case, we still needed to check the current selection and compare it to the visibility filter settings before setting state.  So these methods were added both to the view, and also here in the reducers.  In the end, it really felt bloated and confusing to me, so I rolled back to this.

I think It is reasonable to track these kinds view states the global state tree. Perhaps we want a special state branch for the UI, so that it isn't intermixed with the model state.

Changes like whether the current selection is empty seems like state information that may effect several view components. In this case, it impacts whether to show the 'show only selected' button, and possibly whether or not to show questions.

Including the same complex filter logic on questions in several places seems error prone to me, and not  intention revealing. The button in our example would have to know about the internal representation of questions to make its decision about whether or not to enable itself (based on looking at properties of all the questions).  If we change what this property is called, or how it is stored, we will have to find all the view components which are filtering based on those properties.

knowuh:
(BTW I did merge the other printing issues branch into this because they were both small commits, and while not related exactly,  easier for me to track in one branch)

pjanik:
Well, I don't want to be annoying and block this PR, it's a small change, feel free to merge it, but I'd say I have a different opinion here. :wink: It's not just about this tiny variable, but about the whole pattern.

I believe that it's pretty important not to keep redundant variables in the state. Someone will forget to recalculate them sooner or later. You need to remember to call `setNoSelection(state)` in various actions handlers and that's pretty error-prone IMHO. Especially when we have more functions like `setSomethingInStateBasedOnState(state)`.

Also, I'm not sure what you mean by "view" or "model" state. I'd say there's just one state driving the view, we're working on UI.

Personally, I would just create `noSelection(state)` function that looks like this:
```javascript
function noSelection(reportState) {
  return reportState.get('questions').filter((question, id) => question.get('selected') === true)
}
```
which is just a pure, reusable function. We could put it in `core/no-selection.js` (core sounds a bit serious, but I didn't have better idea). Why would it be confusing to use it both in reducer and in some components? Or why would it be error prone?

[ Actually, we need that in reducer only because we calculate other redundant variable, question visibility, but that's more difficult to fix. Usually this kind of function (calculate something form state) will be necessary only in containers and the result can be later passed down to components as a property. ]

Of course there's a risk that someone doesn't reuse it, but just calculate that again from scratch. But that's pretty much the same risk that someone doesn't use `'noSelection' `state variable, but calculate it from scratch or create another `noQuestionSelected` state variable. And I feel it's still less likely than forgetting to add `setSomethingInStateBasedOnState(state)` in various action handlers. When you need information about selection, you look for it (and hopefully find this helper). When you add a new action handler, you don't necessarily want to review all the unrelated state that might be affected and think what needs to be recalculated.

knowuh:
@pjanik please have a look at 17ea084

pjanik:
It looks good to me, thanks. :+1: And as I said, I'm definitely not like "this has to be done THAT way". It could be merged in any form (as it's a tiny thing), I hope you didn't change it only because I was pushing for it. :wink: I was treating it more as the discussion about (future) patterns. If you still feel confused about it, we can get back to it next time when there's a similar problem.

PR #10, opened 2016-03-28
Change question numbering in reports.

knowuh:
@pjanik  This is based on other PR, so only a17eafb needs review?


pjanik:
@knowuh,
> This is based on other PR, so only a17eafb needs review?

Looks good. :+1: I would also remove number prop (and setting one level higher) if we get a chance, as it seems to be unused now.

knowuh:
I had also missed the same index being used in the student report view.  Should be all set now.

PR #11, opened 2016-03-30
Fix: Print for student report not showing images of image questions.

knowuh:
This problem was intermittant, and the result of images not being (consistently) loaded on
the student page shortly after the print button was pressed.

After taking a deep dive into using image onLoad callbacks &etc, I decided it was simpler to hide the student view (instead of not rendering), so the images would be pre-loaded for printing.  

There may be performance issues with this down the road if we have a lot of student data, but I am not very worried at the moment.

NB: There is an issue with firefox not firing the media query events after the print dialog is closed,
so that we remain on the students report page.  I put this in as an issue ( #12 )

[#115800099]
https://www.pivotaltracker.com/story/show/115800099

### DIFF
--- a/css/report.less
+++ b/css/report.less
@@ -49,5 +49,8 @@
         font-weight: bold;
       }
     }
+    &.hidden {
+      display: none;
+    }
   }
 }
--- a/js/components/report.js
+++ b/js/components/report.js
@@ -16,8 +16,9 @@ export default class Report extends Component {
 
   renderClassReport() {
     const { report } = this.props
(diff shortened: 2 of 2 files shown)


knowuh:
@pjanik take a look if you get a second?  Small change (but I went very far afield and came back.. :) )

pjanik:
Looks good. :+1: I also don't know how to handle that using events. I remember each time I was trying to wait for images in the past, it was failing anyway.

PR #14, opened 2016-03-31
The report should handle case where no students have yet started the activity

knowuh:
This was throwing `transform-json-response.js:86Uncaught (in promise) TypeError: Cannot convert undefined or null to object(…)`

```
// Keys are generated by normalizr above, so just copy them to the object hashes.
function copyAnswerKeysToObjects(answers) {
  Object.keys(answers).forEach(key => {
    answers[key].key = key
  })
}
```
[#116249081]

https://www.pivotaltracker.com/story/show/116249081


knowuh:
Very simple change @pjanik @dougmartin

PR #15, opened 2016-03-31
Fix / workaround an issue with react-responsive-carousel

pjanik:


knowuh:
Looks good. Thanks for the quick fix.

PR #17, opened 2016-04-01
Added sticky headers to class report [#115343939]

dougmartin:


dougmartin:

css/question.less
```diff
@@ -18,6 +18,15 @@
     }
   }
 
+  .sticky-question-header {
+    top: 10px !important;
```
pjanik: I guess you've tried to workaround the fact that react-sticky [doesn't support margins](https://github.com/captivationsoftware/react-sticky#overview--basic-example). However, I'd say it doesn't work too nicely for me in practice. The header is jumping and feels a bit weird. If I remove that style and top margin from `.question-header`, then it's smooth and nice.
So, I guess it would be better to change layout a bit, so the header doesn't require any margin styles (probably remove padding from `.question` and add some question-content container with padding instead).

js/components/question-for-class.js
```diff
@@ -32,22 +34,30 @@ export default class QuestionForClass extends Component {
...
-        <div className='question-header'>
-          <input type='checkbox' checked={question.get('selected')} onChange={this.handleCheckboxChange}/>
-          Question #{question.get('questionNumber')}
-          <a className='answers-toggle' onClick={this.toggleAnswersVisibility}>
-            {answersVisible ? 'Hide responses' : 'Show responses'}
-          </a>
+      <StickyContainer>
+        <div className={`question ${question.get('visible') ? '' : 'hidden'}`}>
+          <Sticky className="sticky-question-header" onStickyStateChange={this.handleStickyStateChange.bind(this)}>
+            <div className={`question-header ${isSticky ? 'stuck-question-header' : ''}`}>
```
pjanik: `<Sticky>` component automatically receives `.sticky` CSS class when it's stuck. So, `.stuck-question-header` could be probably replaced with `.sticky .question-header` in CSS and no state handing would is necessary. Not a big deal, but I'm assuming that eventually we'll also add sticky page and activity headers, so it could simplify a few components.

js/components/question-for-class.js
```diff
@@ -32,22 +34,30 @@ export default class QuestionForClass extends Component {
...
-      <div className={`question ${question.get('visible') ? '' : 'hidden'}`}>
-        <div className='question-header'>
-          <input type='checkbox' checked={question.get('selected')} onChange={this.handleCheckboxChange}/>
-          Question #{question.get('questionNumber')}
-          <a className='answers-toggle' onClick={this.toggleAnswersVisibility}>
-            {answersVisible ? 'Hide responses' : 'Show responses'}
-          </a>
+      <StickyContainer>
+        <div className={`question ${question.get('visible') ? '' : 'hidden'}`}>
+          <Sticky className="sticky-question-header" onStickyStateChange={this.handleStickyStateChange.bind(this)}>
```
pjanik: I'd rather bind handlers in the constructor, just once in the component lifetime. If we call `.bind()` each time we render the parent component, it means we will always re-render this subcomponent, even if it uses [pureRenderMixin](https://facebook.github.io/react/docs/pure-render-mixin.html) (as `.bind` returns a new function each time it's called, so properties comparison will detect a change).
Not too important here (we don't know if `Sticky` is using this optimization or not, probably not), although generally I'd stick to this pattern since all our components use this mixin.

pjanik:
I knew that Dan wanted that and was curious how to implement that nicely in React, so I've taken a closer look. This library and approach looks very good. :+1: I've just left a few inline comments.

dougmartin:
@pjanik I fixed the margin/padding issue so it doesn't jump and @knowuh I added the breadcrumbs from the sequence down to the question when the header is stuck.

pjanik:
Thanks! I could also imagine a bit different design where we have a separate sticky header for each report part (investigation, activity, section, page and question). Then the header wouldn't completely disappear as you scroll down like it does now (since it's limited to the question div only). But probably it would be difficult to use this nice, single-line compact form. So, I guess it's a matter of preferred design. 

Anyway, the code looks good to me and can be merged. :+1:

PR #18, opened 2016-04-01
Add a clear way to close the enlarged image window [#116559811]

dougmartin:
Since the React modal component has no click handler for the image itself I had to mount a click handler on the window when the component mounts.


dougmartin:

knowuh:
Looks good :+1:

PR #21, opened 2016-04-04
Selection status for report filters saves when the checkbox is changed [#116249635]

dougmartin:
NB: The questions in the visibility_filter are now only set when the checkbox is changed.


dougmartin:

pjanik:
Looks good. :+1:

PR #22, opened 2016-04-05
The teacher's individual student report is showing in the new format.

knowuh:
This allows the teacher to click on the student's name in the portal to see a report for just that student.

See corresponding change in concord-consortium/rigse 3c05946

[#115801667]
https://www.pivotaltracker.com/n/projects/736901/stories/115801667

@pjanik @dougmartin  if you have a chance to look.


dougmartin:
Looks good.  The only thing I could see is maybe exporting the "student" and "class" strings as constants to guard against misspellings - I'm not sure if Babel would flag misspelled constants though so that might be a moot point.

dougmartin:
@knowuh Should this be merged or should it wait and be merged with the corresponding commit with rigse?

PR #23, opened 2016-04-06
Changed sticky header z-index so that it shows up behind selected carousel images

dougmartin:


dougmartin:
I'm going to commit this without review as it is just a single css z-index change.

PR #25, opened 2016-04-11
Student reports

knowuh:
This PR incorporates a few small unrelated changes.

* The report can be used to display information to students.  In this case, the report filters and other controls are hidden.

* The print button changes to "Print" when it's a student report.

* The report title changes to "Report for <username>" when its a student report.

* Slight refactoring of the check boxes for selecting questions.

Related to https://www.pivotaltracker.com/story/show/117067031 https://www.pivotaltracker.com/story/show/114578779

@pjanik can you have a look?


js/components/selection-checkbox.js
```diff
@@ -0,0 +1,26 @@
...
+    const {questionKey, onSelectChange} = this.props
+    onSelectChange(questionKey, evt.target.checked)
+  }
+
+  render() {
+    const { selected, hideControls } = this.props
+    if (hideControls) {
+      return null
+    }
+    return(<input type='checkbox' checked={selected} onChange={this.onChange.bind(this)}/>)
```
pjanik: FYI: I think parenthesis expression is only necessary if the opening tag is the next line after `return` statement. 
```javascript
return <div> ... </div>
```
```javascript
return <div> 
... 
</div>
```
but:
```javascript
return (
  <div> 
  ...
  </div>
)
```

js/components/selection-checkbox.js
```diff
@@ -0,0 +1,26 @@
+import React, { Component } from 'react'
+import pureRender from 'pure-render-decorator'
+import Answer from './answer'
+
+import '../../css/question.less'
```
pjanik: I guess Answer and question.less imports are unnecessary.

js/containers/selection-checkbox.js
```diff
@@ -0,0 +1,20 @@
+import { connect } from 'react-redux'
+import { setQuestionSelected } from '../actions'
+import SelectionCheckBox from '../components/selection-checkbox'
+
+const mapDispatchToProps = (dispatch) => {
+  return {
+    onSelectChange: (key, selected) => {
+      dispatch(setQuestionSelected(key, selected))
+    }
```
pjanik: I've been writing actions mapping in the same way, although I think it can be simplified if the action has the same interface as the final property. I think this should work too:
```javascript
const mapDispatchToProps = { setQuestionSelected }
```
or even:
```javascript
import * as actions from '../actions'
(...)
connect(mapStateToProps, actions)(SelectionCheckBox)
```
it's popular pattern, but that will pass all the actions to the container (which is probably okay too).
It's not worth changing now, but we can start using simpler pattern in the future.

js/components/selection-checkbox.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+
+  onChange(evt) {
+    const {questionKey, onSelectChange} = this.props
+    onSelectChange(questionKey, evt.target.checked)
+  }
+
+  render() {
+    const { selected, hideControls } = this.props
```
pjanik: Sorry for my nitpicking, but it would be great to have consistent whitespace pattern in one project (e.g. line 15 vs 20). It's the same about `variable=1` vs `variable = 1` or `[1,2,3]` vs `[1, 2, 3]` etc. I don't really care which ones we use, as long as we use the same pattern within the single project (although leaning towards the most common styles). :wink:

js/data/report.json
```diff
@@ -534,5 +534,6 @@
     ]
   },
   "anonymous_report": false,
-  "is_offering_external": true
+  "is_offering_external": true,
+  "hideControls": false
```
pjanik: I guess we'll receive this prop from Portal as `hide_controls`, so I would make it consistent, as this file is somewhat documenting expected API too.

pjanik:
Looks good. :+1: I've left a few pedantic comments, more as general comments, nothing really needs to be modified now, so feel free to merge.

knowuh:
Thank you for the thorough feedback. I will take a moment now to address some of these items. The slimmer action dispatching really appeals for example.

PR #26, opened 2016-04-19
Redo sticky navigation header.

knowuh:
This PR changes sticky behavior to match the mockups.

Replace `react-sticky`  with the simpler `react-stickynode`

Style adjustments accordingly. 

[#115343939]
https://www.pivotaltracker.com/story/show/115343939

### DIFF
--- a/js/components/activity.js
+++ b/js/components/activity.js
@@ -1,17 +1,18 @@
 import React, { Component } from 'react'
 import pureRender from 'pure-render-decorator'
 import Section from './section'
-
+import Sticky from 'react-stickynode';
 import '../../css/activity.less'
 
-@pureRender
 export default class Activity extends Component {
   render() {
     const { activity, reportFor, investigationName } = this.props
     const activityName = activity.get('name')
     return (
       <div className={`activity ${activity.get('visible') ? '' : 'hidden'}`}>
-        <h3>{activityName}</h3>
+        <Sticky top={60}>
+          <h3>{activityName}</h3>
+        </Sticky>
         <div>
           {activity.get('children').map(s => <Section key={s.get('id')} section={s} reportFor={reportFor} investigationName={investigationName} activityName={activityName} />)}
         </div>
--- a/css/button.less
+++ b/css/button.less
@@ -20,4 +20,7 @@
     cursor: default;
     opacity: .5 !important;
   }
+  &:hover {
+    opacity: .8
+  }
 }
--- a/css/question.less
+++ b/css/question.less
@@ -18,11 +18,6 @@
     }
   }
 
-  .sticky {
-    background: #eb8723 !important;
-    z-index: 1020;
-  }
-
   .question-header {
     background: #8cbbb8;
     color: #fff;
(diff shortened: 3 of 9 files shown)


knowuh:
@pjanik   If you could review these two PR's I would appreciate it.

I found that had to remove `pure-render` annotations for components that had sticky behaviors for this to work as expected.  I am not completely surprised. I was hoping that someone might have written a redux sticky handler that dispatched the appropriate events, but I couldn't find such a thing.

pjanik:
Generally it looks good. I definitely like this approach more, since we don't need to pass all the names down to the question component. And the library looks simpler indeed which is great. 👍 

1. I would cleanup unused properties passed down to the question component which were used by the old implementation. And there's some commented out code which probably isn't necessary anymore, since this implementation looks like the mocked version. 

2. There's a tiny design issue. Report sticky header shows up as the last one, after sequence, activity and page headers.
screen shot 2016-04-20 at 15 05 08 I think it could be easily fixed if we make sure that the report header has height equal to the offset of the sequence sticky header. Offset is 40px, but the report header is 27px and those missing 13 pixels probably cause that.

3. I don't really understand issues with @pureRender. I can't really imagine how that could affect sticky headers. Did you investigate that and have any ideas? @pureRender isn't related to Redux, it just provides `shouldComponentUpdate` implementation that prevents us from unnecessary component re-renders. Actually, I've tried to restore those decorators and things still work fine. How could I replicate those issues?

BTW,
> I was hoping that someone might have written a redux sticky handler that dispatched the appropriate events, but I couldn't find such a thing.

I'm betting that such thing would never exist. Not sure what you mean by redux events, but I don't think there can be anything like "Redux Component". Note that we're mostly dealing with simple React components. Redux layer is just one level higher than React app / components hierarchy. It simply takes care of the state handling and actions dispatching, that's all. It could also be implemented using top-level React component. In my understanding it's purely application layer, not really related to reusable components.

js/components/activity.js
```diff
@@ -1,17 +1,18 @@
 import React, { Component } from 'react'
 import pureRender from 'pure-render-decorator'
 import Section from './section'
-
+import Sticky from 'react-stickynode';
```
pjanik: lint error: unnecessary semicolon :wink:

knowuh:
Thank you for the close review @pjanik  -- I thoguht I saw console error messages  that I had attributed to  pure-render. This:

> If your React component's render function is "pure" (in other words, it renders the same result given the same props and state), you can use this mixin for a performance boost in some cases.

gave me the impression that adjustments to the dom outside of the component definition might violate pure-render expectations.  If I can't reproduce those errors I will add those annotations back.

pjanik:
Good point. I think it's a reasonable assumption, but probably a bit too strict. This decorator / mixin compares new and old props and state, and tells component not to render again if the shallow equality test passes. Probably you can imagine what kind of components would fail. I'd say 99% of components should actually work (as "purity" is built into React design), unless they do some really weird things (it's even hard for me to provide reasonable example) or shallow compare is not enough to detect state changes. Probably this is more common case and components are not pure by default just to be safe. Immutable structures solve that issue.
I feel that even components that manipulate or depend on DOM probably can be pure, as pretty much all the time we assume that component renders the same result for given properties and state (even if there's some "additional state", like the current scroll position).

PR #27, opened 2016-04-19
Slide carousel adjustments

knowuh:
This PR Should be reviewed after #26 

This adds larger fields for the author and notes in image questions.  As per https://www.pivotaltracker.com/story/show/116559551

### DIFF
--- a/css/button.less
+++ b/css/button.less
@@ -20,4 +20,7 @@
     cursor: default;
     opacity: .5 !important;
   }
+  &:hover {
+    opacity: .8
+  }
 }
--- a/css/image-question-details.less
+++ b/css/image-question-details.less
@@ -20,23 +20,48 @@
       width: 50px;
     }
 
-    .slide {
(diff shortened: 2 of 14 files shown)


pjanik:
I'm looking at the second commit only - it looks good. 👍 
Not related to this particular code, but seeing px, pt and ems in styles made me thing if it would be useful to come up with some convention / pattern where and when we use given units. Sometimes I try to stick to 'em' to make sure that most of the content can be scaled up / down just by changing top-level font-size (ofc not being naive, some fixes are always necessary), but I don't remember if I tried to implement it in this repository.

knowuh:
Thank you @pjanik -- the CSS units issue you bring up is a good point, worth more discussion.  I am going to merge this work now so that I can demo.

Have a look at a94f03a  -- I re-added the pure-render annotations, and got rid of all the unused properties / comments.  Thanks for pointing those out.

PR #29, opened 2016-08-19
Add Feedback to portal report.

knowuh:
@pjanik  I hate to do this to you. This is a large pull request, and  I would appreciate a review if you get a chance.

It includes:
- New feedback features.
- Update to react 15
- Report version in API
- Concept of data migration (for backwards portal compatibility)

Some large issue I am aware of and had to punt on:
- I added a reducer for the feedback.  It handles `RECEIVE_DATA` which is redundant. I am sure we could think of a better way to do that.
- There are no new tests.
- Feedback saving doesn't handle network errors at all.
- Components aren't as well factored as I would like.


js/containers/feedback-panel.js
```diff
@@ -0,0 +1,234 @@
...
+    const maxScore = question.get('maxScore')
+
+    const showGettingStarted = (!scoreEnabled) && (!feedbackEnabled)
+    const studentsPulldown = filteredAnswers.map( (a) => {
+      return {
+        realName: a.getIn(['student', 'realName']),
+        id: a.getIn(['student','id']),
+        answer: a,
+      }
+    })
```
pjanik: FYI: I think arrow functions that only return something can be written without return statement and { }, in this case:
```
.map(a => {realName: ..., id: ..., answer: ...}
```

js/core/migrations.js
```diff
@@ -0,0 +1,78 @@
...
+  notify(`migrating to ${migration.version}`)
+  setVersionNumber(dataMap, migration.version)
+  for(migration of migration.migrations) {
+    migration(dataMap)
+  }
+  return dataMap
+}
+
+function skipMigration(migration) {
+  notify(`skipping migration: ${migration.version}`)
```
pjanik: I always see some migrations that are skipped in my JS console (since fake data is v1.0.1) - not sure if it's desired or just old debug code.

js/components/feedback-box.js
```diff
@@ -0,0 +1,42 @@
...
+    super(props)
+    this.state = {
+      value: props.initialFeedback || ""
+    }
+    this.updateText = this.updateText.bind(this)
+  }
+
+  clearUpdateTimer() {
+    if(this.updateTimer) {
+      clearTimeout(this.updateTimer)
```
pjanik: It seems you never clean `this.updateTimer` value. But it also seems that browser don't care if the value really reflects some existing timeout ID.

pjanik:
I've been playing with this feature in practice and skimming through the code too. Not sure what to do about it, as it's really huge PR. I don't think it makes sense to read every single component line by line and comment on it, as it's too much code, probably most of the comments wouldn't be that important and I'm betting there's no time to improve details anyway. So, I was mostly looking at migrations, actions, reducers, etc.

First, I guess it would be nice to cleanup all the warnings and exceptions. I guess it's easier to do it now rather than trying to investigate it later. Some examples:

screen shot 2016-08-23 at 14 32 05
Not sure if it's fixable. Is it related to babel?

screen shot 2016-08-23 at 14 32 12
I think this kind of warning happens when the initial value is undefined or null (empty string should fixit AFAIR).

screen shot 2016-08-23 at 14 32 28
Never seen it before.

screen shot 2016-08-23 at 14 32 54
Also, don't remember it, but seems easy to fix. 

screen shot 2016-08-23 at 14 33 09

I'm not surprised that requests fail when I play with the report locally and use fake data, but I'm a bit surprised that the path is localhost:8080/undefined. I'm not sure if it's related to this PR or not, it doesn't seem so. Probably not important if it works when the endpoint is configured properly.

> I added a reducer for the feedback. It handles RECEIVE_DATA which is redundant. I am sure we could think of a better way to do that.

So, the main question is whether we should keep that in a separate state and reducer. Thinking only about data, I'm not sure as feedback is provided deep in the report JSON structure. So it feels similar to question and answer data which is in the report state. But obviously is kind of a separate functionality, so I agree it makes sense to extract it. 

Maybe instead of creating a separate state and top-level reducer, we could use report state and just create helper functions within report reducer (they could be moved to a separate files too). So, things would be divided in a similar way, but just one step later. Similar helpers (and new files) could be created to handle actions related to questions, answers, compare view etc. We would be dividing reducer just one level lower. Then we wouldn't need to duplicate RECEIVE_STATE handling and all the report data would be together. Or split report data and reducer similarly like you did with feedbacks, but find a good pattern for it. Not sure about it, just thinking loudly. Maybe it would be nice to think about it when we have time to work on https://github.com/concord-consortium/portal-report/issues/4.

js/core/migrations.js
knowuh: I thought it might be helpful to leave in this log message, to track down any potential issues in the future...  Easy to remove though.

js/components/feedback-box.js
knowuh: Good catch, thanks.

knowuh:
Thank you @pjanik  -- This is a huge PR.  Getting rid of as many of those errors as possible seems like a good start.  I think some of these are from the React 15 update. Maybe one is related to `css-transition-group` plugin. Most of the others seem like they would be quick.

The bigger issue about how to handle the reducers is more interesting, important, and confusing. I flip-flopped on that a few times.  I think the current state of the feedbacks reducer isn't great, because of the `RECEIVE_STATE` action as you say.  I agree that we should look at this larger reducer refactoring as part of #4 .

knowuh:
@pjanik  if you get a chance, have a look at the above commits and / or check out the code.  You should `rm -rf ./node_modules` and npm install first.  Most of the changes were updates to node packages, and other minor things.

pjanik:
@knowuh, I've looked at new commits and checked out new code. They look good and report work nicely - no warnings anymore. 👍

PR #31, opened 2017-03-12
Add support of InteractiveState answer type

pjanik:
This PR adds support of `InteractiveState` answer type. When this kind of answer is found, an iframe will be initialized using iframe-phone `initInteractive` call. `ExternalLinkUrl` is still supported. Both answer types work in a very similar way - "View work" link is shown. When it's clicked, an iframe is displayed.

Example:
screen shot 2017-03-12 at 19 59 49

This PR is related to https://github.com/concord-consortium/rigse/pull/303.

### DIFF
--- a/js/components/iframe-answer.js
+++ b/js/components/iframe-answer.js
@@ -1,4 +1,5 @@
 import React, { PureComponent } from 'react'
+import InteractiveIframe from './interactive-iframe'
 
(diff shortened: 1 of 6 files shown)


pjanik:
I've updated this PR to include one more change - interactive names are displayed next to the question number:

screen shot 2017-03-15 at 19 26 30

PR #32, opened 2017-09-22
Add link icons

knowuh:
This story asks for a clickable link in the report that takes the teacher to the activity page for the current answers:

https://www.pivotaltracker.com/story/show/151223803

In our discussion, we decided the above feature was complete, but it needed a better visual indicator.


.travis.yml
```diff
@@ -2,11 +2,12 @@ language:
...
   - ruby
 node_js:
-  - 5
+  - 6
 rvm:
   - 2.2.0
 install:
   - travis_retry gem install s3_website -v 2.12.3
+  - travis_retry yarn
   - travis_retry npm install
```
sfentress: What's the purpose in running both `yarn` and `npm install`? Will `npm install` change the package structure after `yarn` has already run?

knowuh: Thanks for catching this -- just an oversight.

PR #35, opened 2018-02-09
Fix labbook items

scytacki:
This fixes a problem with some recent work in portal-report.
The result of that work was that interactives that should show as embedded iframes in the report were not showing up at all.  Clicking a view work link would have no effect in this case.

The changes in this PR are:
- do not automatically show iframes when opening the feedback dialog. This is for performance reasons. If a there are 30 students in a class and they all have done an interactive that wants to be reported on as an embedded iframe, this would really slow down the page browser when the dialog is opened.  This change is just the removal of the `alwaysOpen` from the answer component in the feedback-row.
- fix issue with logic about showing the embedded iframe. The problem with the old logic is that it was looking at `props.displayInIframe` which doesn't exist, it should have been looking at `props.answer.get('displayInIframe')`
- add an external-link decorator to 'view work' links that open in a new tab.
- handle the old style of json that would be returned by an older version of the portal
- add new example questions to the included report.json to make it easier to test this without connecting it to a portal

You can test out the running code here:
https://portal-report.concord.org/branch/fix-labbook-items/

### DIFF
--- a/js/components/feedback-row.js
+++ b/js/components/feedback-row.js
@@ -97,7 +97,7 @@ export default class FeedbackRow extends PureComponent {
       <div className="feedback-row">
         <div className="student-answer">
           <h3>{name}'s Answer</h3>
-          <Answer answer={answer} alwaysOpen={true} />
+          <Answer answer={answer} />
         </div>
         { answered ? this.renderFeedbackSection(answer) : null }
       </div>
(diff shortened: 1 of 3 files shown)


scytacki:
Yes report.json was copied from some code generated by learn.staging.concord.org for a few different activities.  I don't think it needs more scrutiny since is just for developers and testers to be able to test out things more quickly.

PR #36, opened 2018-03-02
Auto-scoring bug fixes

knowuh:
Trying to present a smaller PR this time ;)

These are related to this auto-scoring story:
https://www.pivotaltracker.com/story/show/151224400

It fixes a bug where a student's activity level `automatic score` were summing question scores that included questions that were no longer being scored.  (Teacher changed their mind but we save the students scores incase the teacher decides to re-enable scoring)

Now when computing automatic scores for the activity or calculating the `maxScore` we check `scoreEnabled` first.

Also a fix for 'NaN' in a score field for `parseInt('')`,  and a change in the default `report.json` to show auto scoring working for an activity.

### DIFF
--- a/js/core/activity-feedback-data.js
+++ b/js/core/activity-feedback-data.js
@@ -82,7 +82,9 @@ export function calculateStudentScores(state, questions) {
   }
 
 
-  const scores = questions.map( q => q.get('answers'))
+  const scores = questions
+    .filter(question => question.get('scoreEnabled'))
+    .map( q => q.get('answers'))
     .flatten()
     .map(answerId => report.getIn(['answers', answerId]))
     .groupBy(answer => answer.get('studentId'))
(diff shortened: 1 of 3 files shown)


knowuh:
@pjanik  -- I forgot to tag you :)

js/core/activity-feedback-data.js
```diff
@@ -97,8 +99,9 @@ export function calculateStudentScores(state, questions) {
 
 export function getComputedMaxScore(questions) {
   return questions
+    .filter(question => question.get('scoreEnabled'))
     .map(question  => question.get('maxScore') || 0)
-    .reduce( (total, score) => total + score)
+    .reduce( (total, score) => total + score) || 0
```
pjanik: Just wondering which case is this `|| 0` handling? If scores are all 0, it should return 0 anyway. If we call `reduce` on an empty array (e.g. due to filtering above), `reduce` will throw an error if we skip its second argument (initial value). If that was your intention, we'd rather need to use `.reduce( (total, score) => total + score), 0 )`. It handles empty array case.

knowuh: Thanks for this. I will make this change.

PR #37, opened 2018-03-07
Beginning to add Rubric features.

knowuh:
## PR to start epic about adding Rubric scores to student feedbacks.

The summary of this feature is that activity authors can create criteria for evaluating students work at  the activity level. The rubric is a matrix of grading categories and proficiency levels. 
 
There are several rubric related stories tagged with `feedback` and PT.  This PR handles these two stories:

https://www.pivotaltracker.com/story/show/155404185 
https://www.pivotaltracker.com/story/show/155404153

> A teacher can select "use rubric" as one of the activity level feedback options, and they will see the rubric for this activity appear in activity-level feedback.

> An author can define a rubric using JSON. This structure of the rubric would be any number of grading categories, which could be scored using any number of proficiency levels. There could be a numerical value associated with a proficiency level.  … (more in story)

Before I started working on these features, I extracted `FeedbackOptions` from `FeedbackOverview` in e65dca3

Most of the view related changes happened in deb893d

Actions and reducers were added in 1d19b73

## Notes:

The rubric is loaded asynchronously from URLs.  We imagine that rubric will eventually be hosted in the portal, but for MVP purposes we might host rubric elsewhere.  

This was my first attempt to work with asynchronously updating the state tree, and I tried to follow the patterns set out in the ApiMiddleware, but eventually had to learn about dispatching actions from within actions using Thunk.

I feel good about the implementation, but it was new to me.  

I found that in order to use thunk within the ApiMiddleware I had to change the middleware load order.

Thanks for looking at this long PR @pjanik

### DIFF
--- a/js/actions/index.js
+++ b/js/actions/index.js
@@ -1,6 +1,6 @@
 import fetch from 'isomorphic-fetch'
 import { fetchReportData } from '../api'
-
+import { requestRubric } from "./load-rubric";
 export const INVALIDATE_DATA = 'INVALIDATE_DATA'
 export const REQUEST_DATA = 'REQUEST_DATA'
 export const RECEIVE_DATA = 'RECEIVE_DATA'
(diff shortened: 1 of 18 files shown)


js/actions/index.js
```diff
@@ -23,22 +23,27 @@ export const ENABLE_ACTIVITY_FEEDBACK = 'ENABLE_ACTIVITY_FEEDBACK'
...
-    }
+  return (dispatch, getState) => {
+    dispatch({
+      type: REQUEST_DATA,
+      callAPI: {
+        type: 'fetchReportData',
+        successAction: receiveData,
+        errorAction: receiveError
+      }
+    });
```
pjanik: Is that necessary? It's been a while since I used Redux last time, but I think explicit dispatch call (or thunk middleware) is useful either when you have some async operation, multiple actions dispatched in one function, or you need to use `getState`. This action looks kind of straightforward.

knowuh: You are right, this is not necessary.  I am not sure what happened here.  I might have just been going for consistency.

PR #38, opened 2018-03-13
RubricFeedback for students #155408322

knowuh:
For this PT Story:

"A teacher can select levels of proficiency in rubric-based feedback for each criteria and that score will be saved for that student."

https://www.pivotaltracker.com/story/show/155408322

This PR modifies actions that send activity-level feedback to portal report API, to include the students rubric feedback. Changes are in actions/index.js `enableActivityFeedback` and `updateActivityFeedback`

Corresponding view changes, and updates to the sample data. `report.json` and a new file `public/sample-rubric.json`

PR #39, opened 2018-03-15
155368181 Freeze rubric into offering feedback

knowuh:
@pjanik I think you are looking at #38 -- That should be reviewed before this.   I could call upon @scytacki if you need a break, since he is looking at the portal piece of this same story.  ( rigse/#447 )  

( This PR will be much smaller than #38 once that has been merged )

----

Feature: Save & Load rubric (content) using report API.

We now 'Freeze` rubric after it has been loaded from a URL. This is to prevent rubric from changing after a teacher has
graded some or all of their students.

General process:

1. When report data is first loaded from the server, we load rubric for activity feedback, if it exists in serialized form.
2. After the report data has all been loaded, we check for missing activityFeedback rubric for which we have a URL.
3. If the URL exists, but we haven't loaded a copy of the rubric yet, we load it.
4. We update any activities that don't have a serialized version of the rubric JSON using the portal `enableActivityFeedback` api

Main changes:
1. The `rubricReducer` now handles `RECEIVE_DATA` actions. It looks for serialized rubric. If it finds some it puts it into the state.
2. The 'addRubric` action now dispatches `enableActivityFeedback` actions for activities that are missing serialized rubric.

PT Story:
If a teacher uses a rubric to give feedback, it is copied and locked for them for that assignment, so that it does not change if the author changes it after that point.

[#155368181]

https://www.pivotaltracker.com/story/show/155368181


js/actions/load-rubric.js
```diff
@@ -48,16 +71,29 @@ export function loadRubric (data) {
...
+      .forEach((u) => {
+        dispatch({
+          type: REQUEST_RUBRIC,
+          callAPI: {
+            type: 'loadRubric',
+            data: u,
+            successAction: addRubric
+          }
+        })
+      })
```
pjanik: Those two API calls, `withRubric` and `needingFetch` are a bit unclear to me. I guess if I analyzed everything better, I could understand it, but maybe some comments could be useful. It's kind of confusing, as one action is LOAD_RUBRIC, other is REQUEST_RUBRIC, which is in fact API call "loadRubric", and so on. 😉

knowuh: Yes – this is a very good point.  Its not obvious what is happening here. It needs some explanation.

Thanks!

PR #40, opened 2018-03-21
Improvements to student feedback

knowuh:
This pull request improves the student views (and the teacher's print view) in two key ways:

1. It adds feedback from the rubric to the activity level view.
2. It unifies all feedback areas with a consistent look. It also aligns feedback scores so they always appear on the right hand side, allowing teachers and students to easily check scores.

This PR is mainly component / container changes.  I also changed the dummy report JSON in minor ways to excise the new views.

PT: Story

[#153265633]
The student can see activity level feedback at the top of the report on their work.

When teacher does not have a text feedback, the score appears on the left side. When there is a text feedback, the scores are on the right side.

It would be easier to glance down one side to see scores if they were all on the same side. See images

https://www.pivotaltracker.com/story/show/153265633

### DIFF
--- a/css/activity-feedback-for-student.less
+++ b/css/activity-feedback-for-student.less
@@ -1,0 +1,45 @@
+.activity-feedback {
+  display: flex;
+  justify-content: flex-end;
+  .actFeedbackPanel {
+    padding: 0.5em;
+    border-radius: 0.5em;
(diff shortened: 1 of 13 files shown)


knowuh:
@pjanik if you get a chance, I would appreciate your eyeballs.  👀 

Thanks!

PR #42, opened 2018-03-22
155798025 preview rubric layout

knowuh:
This PR creates a new html page called `rubric-test.html` which can be used by authors to try out rubric JSON.  This is a **quick and dirty hack** to use until a graphical tool can be built.

The original story was to use an URL input field, but it was understood that a text area for writing complete JSON would also be acceptable.

PT Story:

An author writing a rubric can go to a URL to preview the rubric if they provide a parameter that is the URL to the rubric data.

[#155798025]

https://www.pivotaltracker.com/story/show/155798025

### DIFF
--- a/js/containers/rubric-test.js
+++ b/js/containers/rubric-test.js
@@ -1,6 +1,6 @@
 import React, { PureComponent } from 'react'
 import RubricBox from '../components/rubric-box'
-import FeedbackPanelFroStudent from '../components/feedback-panel-for-student'
+import FeedbackPanelForStudent from '../components/feedback-panel-for-student'
 
 const sampleRubric = {
   'id': 'RBK1',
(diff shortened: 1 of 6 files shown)


knowuh:
@kswenson  if you have a moment and could give this a quick scan 👀, I would appreciate it.

js/containers/rubric-test.js
```diff
@@ -0,0 +1,121 @@
+import React, { PureComponent } from 'react'
+import RubricBox from '../components/rubric-box'
+import FeedbackPanelFroStudent from '../components/feedback-panel-for-student'
```
kswenson: Typo: "For"Student

public/rubric-test.html
```diff
@@ -0,0 +1,32 @@
...
+      font-family: Arial, Helvetica, sans-serif;
+      background-color: white;
+      padding: 10px;
+      border-radius: 0.5em;
+    }
+  </style>
+</head>
+<body>
+  <div
+    id="app"></div>
```
kswenson: Don't really need two lines here.

PR #48, opened 2018-04-10
155404750 light box rubric summary

knowuh:
This commit creates a light box view of the rubric summary.

Feature: Lightbox rubric summary view. +fix colors & border.  …
Fix the colors and borders for feedback summary:
[#155404659]
https://www.pivotaltracker.com/story/show/155404659

Create a lightbox rubric feedback summary
[155404750]
https://www.pivotaltracker.com/story/show/155404750

### DIFF
--- a/js/components/rubric-summary.js
+++ b/js/components/rubric-summary.js
@@ -1,0 +1,81 @@
+import React, { PureComponent } from 'react'
+import chroma from 'chroma-js'
+import '../../css/rubric-summary.less'
+import RubricBox from './rubric-box'
+export default class RubricSummary extends PureComponent {
+  constructor (props) {
+    super(props)
+    this.state = {
+      showLightbox: false
+    }
+  }
+
+  openLightBox (event) {
+    console.log('open')
+    this.setState({showLightbox: true})
+  }
+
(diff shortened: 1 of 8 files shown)


js/components/rubric-summary.js
```diff
@@ -0,0 +1,86 @@
...
+    this.setState({showLightbox: false})
+  }
+
+  renderLightBox () {
+    const {rubric} = this.props
+    const rowMaps = this.getRowMaps()
+    return (
+      <div className='rubric-lightbox-wrapper'>
+        <div className='rubric-lightbox'>
+          <div className='closebox' onClick={this.closeLightBox.bind(this)}>✖</div>
```
pjanik: `bind` will return a new function each time you call it. It means that every time you render parent component, this child component will have some property changed and it will propagate react invalidation / re-rendering machinery further down, even if it's not necessary. In most cases it's not very imporant, but if the given child component is expensive to render, it might make some difference (here it's just a DIV, so whatever). Or if we have multiple / hundrets of cases like this (when you render some list). Same with dynamically defined handlers like `onClick={ (e) => doSomething(e) }`.

PR #49, opened 2018-04-13
155408233 autoscore with rubric scores

knowuh:
@pjanik 

I am afraid I am not completely satisfied with this PR, but I have to leave it here for now.  

Because of the multiple ways we can calculate scores, I feel like things have gotten pretty ugly in this commit.  And probably, these different scores should be calculated in a reducer, instead of done in view helpers in `activity-feedback-data.js`

I am sorry I didn't iron this out more.  I am leaving you with something incomplete.


pjanik:
I'll take a look at this PR later next week. It doesn't seem we're in a hurry, as a lot of people are on vacations (including you). And there's a chance that I'll start working with portal-report codebase soon, so it might make review easier and more valuable. At this point, I'm not able to analyze / understand issues you mentioned.

scytacki:
@pjanik and @knowuh can we get this wrapped up?
Noah, as far as I know you are going back to @ddamelin project after the portal demolition work.  
So if Dan is fine with it, then you could balance any time you spend wrapping this up, by coming back to Dan's project later.

Or perhaps we can release the portal-report without this PR?

knowuh:
@scytacki I think the right thing todo is release portal pages as is, without this PR.  But maybe we should double check with Dan.  He is demoing using this branch. We can discuss in the planning / demo meeting at 12:00.

scytacki:
To follow up on this after the meeting. My understanding is that we are going to try to get this done before releasing the portal-report. As far as I know there isn't a reason to rush the release of the portal-report.

knowuh:
This is still a huge PR -- however I think it is a little better organized now, and it is using memoized selectors via `reselect`.  

@pjanik  when you get a chance for a review I would appreciate it.

js/selectors/activity-feedback-selectors.js
```diff
@@ -0,0 +1,293 @@
...
+  MAX_SCORE_DEFAULT,
+  RUBRIC_SCORE
+} from '../util/scoring-constants'
+
+/******************************************************************
+ * These functions compute data for views. Some of the computations
+ * are expensive, so we use reselect to memoize the results. Beacause
+ * we use the selectors on multiple containers within the same
+ * document we use the `makeGet<property>` factory method pattern
+ * to bind each memoized function with the appropriate container.
```
pjanik: I've been using reselect and I've never had to do this additional memoization and `make...` pattern. I guess the reason to do that is not really because we have multiple containers, but because some of the selectors are based on component properties (e.g. activity), right?

If we use just the global, singular Redux state, then obviously result will be the same for every component, so we don't need to worry about the memoization. But if also use component-specific properties, then we do need separate instances for every component. Is my logic correct? As I'm saying, I've never seen this pattern before, so I'm wondering if I understand it correctly.

If so, maybe we could update comments here that this is necessary only if your selector is based on properties. And review selectors if they really need to use this pattern? Or are all of them based on properties? It seems that only `getActivity` is property-based, but also it's quite likely that this activity is passed almost everywhere.

Also, I think I'd find it helpful to group functions into three categories (perhaps just using some comments and order):
- helpers
- simple state-based inputs (selectors)
- complex selectors using other selectors as inputs

I think this is almost like this now, but maybe some clear division could be nice.

I've noticed that sometimes we have only `makeGetSomething` definition, and sometimes we additionally specify `getSomething` as a separate function. I'd lean towards just single `makeGetSomething` definition, so I can clearly see that this code is a selector. Or at least keep them next to each other (I think that's always the case except from `getQuestions`).

knowuh:
@pjanik -- You are right; it is because we are using the `activityID` property in components that we need to create a new memoization for each component.  We partition the data by activity, and so all our nested selectors start by checking the `scoreType` for example of the activity. Since selectors further up the chain need this partitioning its factories all the way down  :)

I added some documentation, including a link to the [reselect documentation section](https://github.com/reduxjs/reselect#sharing-selectors-with-props-across-multiple-component-instances) which describes this pattern. 

There is another library called `re-reselect` which might have made things cleaner for us.  In that case we could define our own cache-keys for the memoization functions, and it would not have looked so strange. I was tempted to use that library, but I thought maybe explicit would be better.

I made some minor changes thanks to your feedback, including organizing code by helper, simple selector, and compound selector.  I also rolled-up the helper methods that really might have been more cleanly expressed directly inside the compound selectors themselves.  Ironically (?) I had put the `getFoo ()` functions outside the `makeGetFoo ()` because I was remembering your advice about not repeatedly creating anonymous functions.

I hope you don't mind having one more look?

pjanik:
Thanks, it was mostly for my own learning process, as I might need to use it soon too. It looks great now. My questions about `makeGetSomething` vs `getSomething` doesn't make sense now when we have only `make...` version. 😉

PR #52, opened 2018-05-01
157206977 Hide activity level feedback options for question level feedback.

knowuh:
Questions were showing options for 'use rubric for scoring' and other things which weren't possible.

This bug was a result of a partially completed factor-out of a common `feedbackOptionsView.` component.

This PR finishes the refactoring.

NOTE: Should be reviewed after #51

This PR if for this [PT story](https://www.pivotaltracker.com/story/show/157206977)


pjanik:
👍

PR #53, opened 2018-05-04
Fix: CSS that was preventing horizontal scrolling.

knowuh:
Also fixes:  Prompt being cut off in activity feedback lightbox.
Also fixes:  Recovers some vertical space by grouping input fields into grid.

Story:

[#156548713]
https://www.pivotaltracker.com/n/projects/736901/stories/156548713

Rubric panel wont display all columns in narrow view

### DIFF
--- a/css/feedback-panel.less
+++ b/css/feedback-panel.less
@@ -10,9 +10,6 @@
 }
 .feedback-container {
   .feedback-panel {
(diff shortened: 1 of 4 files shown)


knowuh:
This is CSS only change, so I will merge myself …

PR #54, opened 2018-05-04
Student sees table feedback for their rubric rating

knowuh:
This is NOT TO BE MERGED.  I am putting the PR in only for documentation purposes.  

It's  incomplete, and should be picked up by the next developer.

In the student report feedback section, if the teacher is using a rubric with a student version, the student sees a table at the top of the page that shows the student version of the rubric for the levels the teacher marked this student at.

This commit includes:
* some documentation in RubricDescriptionFeedback.md
* a helper class for selecting correct feedback
* a test for the helper class
* the extraction of a new (but not completed) react component for the table view.

Stories:
[#157292985] [#157292888]
https://www.pivotaltracker.com/n/projects/736901/stories/157292985
https://www.pivotaltracker.com/n/projects/736901/stories/157292888

# Student rubric feedback will now display in a grid.

A new pure and stateless React component has been started `rubric-box-for-student`. Right now it does not have the correct appearance.  It uses a new class `rubric-helper` to work out what to display.

## Student rubric feedback will display `ratingDescriptions`

Each row in the grid will display the student results for a particular critereon.
The fist column will describe the critereon, and the second will describe the students rating.

In the JSON structure for the rubric defines a `criteria` and `ratings` section.
At the intersection of `criteria[1]` and `rating[2]` is a `ratingDescription`.  
These rating descriptions are nested under the `criteria[1]` in JSON it looks like this:

```
            …
            "ratingDescriptions": {
                "R1": "Student provides reasoning that …",
                "R2": "Studentdescribes predator-prey …",
                "R3": "Student does not provide reasoning …"
            }
            …

```

## Student rubric feedback may display alternate descriptions by using `ratingDescriptions_student`

The above rating descriptions look to 'teacher-centric'.  So if the rubric author so wishes, they can specify feedback which is specifically oriented to the student, using an adjacent field in the JSON called `ratingDescriptions_student` eg:

```json
            …
            "ratingDescriptions": {
                "R1": "Student provides reasoning that …",
                "R2": "Student describes predator-prey …",
                "R3": "Student does not provide reasoning …"
            },
            "ratingDescriptions_student": {
                "R1": "You provided reasoning that …",
                "R2": "You described predator-prey …",
                "R3": "You did not show me your reasoning"
            },
            …

```

Using the `util/rubric_helper.js` class, you can ask for feedback results formatted for presentation to students specifically.

## Tests

There is a `test/rubric-helper_spec.js` mocha test that attempts to test and document the `rubric-helper` class. All the mocha tests can be run continuously by running `npm run test:watch`

## Sample Data

The mocha test, and manual browser test at http://localhost:8080/rubric-test.html use sample rubric data that you can find in  `public/sample-rubric.json` and `public/sample-rubric-feedback.json`

#### Table layout,

As mentioned before the `rubric-box-for-student` is not complete. It just uses the helper and the new description fields. Work needs to continue so that the feedback is displayed in a table. The goals is that each criteria description shows up in the left side, and the students rating description shows up on the right.

#### More student-centric descriptions and labels.

The rubric JSON defines a `description` field for each of the `criteria` elements too. We will want to extend the pattern we are using in the `rubric-helper` to be able to conditionally extract `description_student` from each criteria definition. The pattern of appending "**`_student`**" to field names will provide us a with an obvious naming convention.

Here is an example:

```json
      "criteria": [
          {
            "id": "C1",
            "description": "Student must supported claims by evidence …",
            "description_student": "You must make a claimm, and  support it w/ evidence …",
            "ratingDescriptions": { … },
            "ratingDescriptions_student": { … }
          },
          …
      ]
```


pjanik:
@knowuh, I think it's ready now. It's obviously a continuation of the work you started, but also a few unrelated bits. Some things were broken and I had to fix them. I found some unrelated issues too that I could quickly fix.

1. I removed a few Rubric reducers and left `LOAD_RUBRIC` only. I was asking some time ago about the difference between it and `RECEIVE_DATA`, and now I was able to confirm that - they were both doing the same thing (and causing some issues too). Just in a different way. RECEIVE_DATA was parsing raw JSON response, and LOAD_RUBRIC was triggered by `requestRubric` action that was actually processing Redux state. I decided to leave LOAD_RUBRIC as it's used both to load cached Rubric and one that is dynamically downloaded.

2. I tried to cleanup some Rubric actions and API. I was getting lost in "request", "load", "add" etc., names and api, so I tried to make it more similar to main actions that download report data. Also, added a bit better error handling. If Rubric URL is incorrect, report will still work, but it won't show "Use rubric" option.

3. I had to change naming convention from "_student" to "ForStudent". "_student" was getting lost during automatic camelization of the Portal API JSON response ("_student" -> "Student"), so it didn't work for cached Rubrics.]

4. I've added all the requested features, so student criteria, ratings, criteria label and feedback label. Plus changed tooltips displayed to teachers.

4. Fixed rubric-test.html to use the same "public/sample-rubric.json" file.

5. Fixed getting started overlay.

6. Slightly rectored Activity container, so it was a bit easier to change layout (center feedback).

js/components/rubric-box.js
```diff
@@ -34,10 +34,11 @@ export default class RubricBox extends PureComponent {
     const checked = rubricFeedback &&
       rubricFeedback[critId] &&
       rubricFeedback[critId].id === ratingId
-    const ratingDescription = (
-      crit.ratingDescriptions &&
-      crit.ratingDescriptions[ratingId]
-    ) || null
+    // Tooltips displayed to teacher should actually show student description if it's available.
+    const ratingDescription =
+      (crit.ratingDescriptionsForStudent && crit.ratingDescriptionsForStudent[ratingId]) ||
```
knowuh: Couldn't we use the rubric helper methods here?

js/containers/activity.js
```diff
@@ -22,7 +22,7 @@ import '../../css/activity.less'
 
 class Activity extends PureComponent {
   constructor (props) {
-    super()
+    super(props)
```
knowuh: Wow.  Its weird this worked before.

js/components/rubric-box.js
```diff
@@ -30,13 +31,12 @@ export default class RubricBox extends PureComponent {
...
-    const ratingDescription = (
-      crit.ratingDescriptions &&
-      crit.ratingDescriptions[ratingId]
-    ) || null
+    const checked = !!(rubricFeedback && rubricFeedback[critId] && rubricFeedback[critId].id === ratingId)
+    // Tooltips displayed to teacher should actually show student description if it's available.
+    const ratingDescription =
+      (crit.ratingDescriptionsForStudent && crit.ratingDescriptionsForStudent[ratingId]) ||
+      (crit.ratingDescriptions && crit.ratingDescriptions[ratingId]) ||
+      null
```
knowuh: Couldn't we use the rubric helper here?

pjanik: Not really, as I think it's student-oriented. It's not used in this component at all. It helps with student box (to connect all the ratings with descriptions and so on), but not sure if it's that useful here. If we wanted to extend it, maybe we should just create set of independent functions instead of monolith class that handles all the cases.

If I remember correctly, `RubricHelper` accepts rubric JSON and particular rubric feedback that we don't have here (as we're just rendering table to let teacher create it).

pjanik:
Thanks. I thought about helper there, but it didn't seem that helpful in this particular component (unless we refactored it all). I'm going to merge it then.

PR #55, opened 2018-05-07
Fix:  Overlay panel for 'getting started' could not be dismissed.

knowuh:
Because of a positioning stylesheet regression, the 'getting started' box could not be dismissed.

Story:
[#157362447]
https://www.pivotaltracker.com/story/show/157362447


knowuh:
This is a critical bug with a simple fix, so I am merging it now.

PR #56, opened 2018-05-09
Export some React components as a separate library

pjanik:
Portal Pages needs to use some Rubric-related components.

I've added a way to define separate library entry point that can be easily extended in the future.
I think README should describe how this is supposed to work and be deployed. Let me know if it's not clear.

Portal Report itself will provide two ways to use this lib - NPM and just an old-school script tag.
The question is which one we want to use in Portal Pages. 

I guess script tag could be slightly more automated, e.g. Portal itself could use Portal Report URL to generate library URL and then include it before Portal Pages. So, each time we would update Portal Report, we would update library too. But it seems a bit messy, we're depending on global variables, and so on.

I'd prefer NPM package defined in Portal Pages. It's a bit more work (we need to publish package and then update it in Portal Pages) but seems way more elegant. I think we wouldn't have to do it very often though. Even if the library is out of sync with the Portal Report app, in most cases it shouldn't cause any problems, as the library is only a tiny subset of the code. We can update it only when it's necessary. This is question doesn't really affect this PR as it provides both ways anyway, but perhaps it'd be good if we think about it now, so I can setup other projects in the prefered way (cc @scytacki, @knowuh).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,9 @@ node_js:
 install:
   - travis_retry gem install s3_website -v 3.4.0
   - travis_retry yarn
-before_script: webpack
+before_script:
+  - webpack
+  - WEBPACK_TARGET=lib webpack
 script:
   - "npm run test"
   - ./s3_deploy.sh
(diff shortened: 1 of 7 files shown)


pjanik:
Thanks, I'm merging it then (@scytacki, you might want to take a look at it anyway, as it might slightly affect future release process).

PR #57, opened 2018-05-10
Fix Rubric reducers and tweak Rubric Box

pjanik:
1. Rubric Box: pretty simple visual tweak - when there's data for only one student, use check mark instead of 100% and 0%. It's useful in Portal view where we display this icon and box for every student separately.

2. Rubric reducers - it seems to me that `RECEIVE_DATA` was totally broken. It didn't work for sequences (as they're not supported), it didn't do anything for activities (as it was going through activity children=sections that had no rubric information). In case of Activity, everything was set in LOAD_RUBRIC action anyway. I'm actually adding support of sequences, so I had to fix it, so the map is correct and error isn't thrown in another place. BTW, @knowuh, what's the purpose of this RECEIVE_DATA handler? Do we need it at all? I feel that we don't, as it was broken before and things were working more or less.

### DIFF
--- a/css/rubric-box.less
+++ b/css/rubric-box.less
@@ -23,6 +23,10 @@
         background-color: hsla(0,0%,100%,0.8);
         padding: 0.25em;
         border-radius: 0.25em;
+        min-width: 1em;
+        min-height: 1em;
+        text-align: center;
+        line-height: 1em;
       }
     }
   }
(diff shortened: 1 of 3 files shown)


knowuh:
`RECEIVE_DATA` is called only when we request the report data from the server at set up for the very first time while we are loading the report, and then on schedule. The rubric and rubric_url are sent as part of the Activity JSON we load.  

This rubric reducer is listening to the same event as the main reducer, but loading only the data it needs to maintain its state going forward -- which is the rubric cache basically.

PR #59, opened 2018-05-25
Fix handling of external reporting_url from LARA [#157857724]

dougmartin:


scytacki:
Did you track down why this started happening?

My theory is that it is only a problem if some learner data was saved before the 'has report url' property is set on the interactive. So the portal saves the full answer as the answer. Then when this property is set, the portal doesn't go back and update the data saved as the answer.

So I'd guess that if you have a student that opens the interactive for the first time after the 'has report url' property is set they won't have this problem.

I'd like to understand this better before adding in this fix. It might be that the portal should be changed to reprocess the answer when the question is changed instead of doing it in the browser.

dougmartin:
@scytacki I added your note to the code.  I didn't have anything to add to the note.

PR #60, opened 2018-06-12
Refactor existing portal report, move code into "report" directories/namespace

pjanik:
[#158162285]

I'm opening this PR still without Dashboard code, so it's easier to review. The diff might be difficult to read, so a little summary:
- The report code is moved to `report` directory (or `common`), e.g. `components/activity.js` -> `components/report/activity.js`. In the future, there will be also dashboard subdirectories, e.g. `components/dashboard/activity.js`. I think it useful, as otherwise, we'll have multiple components with very similar names and no clear hint whether they are used by the old report or dashboard.
- I converted a few functions into reselet selectors. Especially `report-data.js`. Now it's called report-tree and it's a selector. I used a "tree" name, because it's converting flat redux state into tree / a hierarchy. It will be used by the Dashboard too.
- Header has been extrated into a separate component (used by Dashboard too, also an example of a component used by two apps). 
- Small style changes. Padding is a bit smaller so the dashboard can use more horizontal space. I changed that in report too, so it's consistent. We'll probably add a button to switch between these two views, so it would look strange if header was chagning its width.

@knowuh, I guess you'd be the best person to review this PR, as you're working with this code and might work on it soon. I think that Dashboard-related PRs can be reviewed by Eli in the future when he gets familiar with the codebase.


pjanik:
> Thanks for the "*-tree" selector naming idea. It makes sense. Its funny that we take hierarchical data from the portal, then normalize it, then re-nest it. But this is a really nicer way to do it.

Right, it seems a bit awkward. Flat structure is useful for updating various models (e.g. a question with given ID, without having to traverse everything from top to bottom). But hierarchical structure works better with rendering usually, as it resembles components hierarchy. Another approach would be that every model accepts just "id" as a property, and it gets data from flat Redux state itself. It would work, and the tree structure wouldn't be necessary, but it means that almost everything would need to be a "container". This is what MobX does in fact. Redux guides were quite often recommending to keep as little containers as possible, but this part is very unclear to me (and worst part of working with Redux, as I feel bad each time I create a new container, but I don't know why hehe).

> Did you consider name-spacing the actions too? I realize its not as important or clean. Just curious about it.

Not yet. Some actions are common, some used by Report only. I guess I'll start adding dashboard-specific actions to a separate file and will think about the old ones.

PR #61, opened 2018-06-13
Add basic Dashboard components and table layout

pjanik:
Adds a basic Dashboard layout. No proper styling yet (just borders to make sure that all the divs are matching correctly). I didn't use `<table>` as its structure doesn't work well with React component hierarchy - activities are columns, but you need to render the whole row at once in the HTML `<table>` element. Instead, I've used a bunch of regular divs.

CSS might look a bit complicated, but I think that LESS variables will be a way to go to keep things reasonably simple and ensure that cells have appropriate dimensions (since we can't use a table that does it for us).

`dashboard.js` component has also some magic related to scrolling and resizing. Since both student names and headers are locked, I had to synchronize one of them with main table area. You can avoid JavaScript if you lock just one of them (e.g. only 1st column or 1st row), but I don't see a way do it when both of them need to be locked.

You can see that behavior in practice if you make the window small enough (our test data isn't very large, so you need to make resize both width and height):
http://portal-report.concord.org/branch/dashboard/index.html?dashboard=true

I added a new reducer and I'll keep adding actions to dashboard-specific file. At some point we could think about refactoring existing actions and reducers into common ones and report-specific. The small problem is that some report-specific state is merged into report data. It is useful when you work on components, but I'll try to avoid it for dashboard if possible.

@knowuh, perhaps it would be the last dashboard-specific PR I'm asking you to review.
@ekosmin, I guess that's the first PR that you can take a look at to see how things are going to be working.


knowuh:
I am still reviewing this.  So far it looks good. 

For the CSS I was wondering if you thought about looking into [CSS Grid layout](https://developer.mozilla.org/en-US/docs/Web/CSS/grid)?   It seems to be widely supported now according to [can I use](https://caniuse.com/#feat=css-grid)

Not sure if it will grid will fit your needs, but I do find the logic in the CSS a bit tricky to follow.

Nice scroll sync!

pjanik:
> For the CSS I was wondering if you thought about looking into CSS Grid layout? It seems to be widely supported now according to can I use

Actually, I didn't know that it existed. Thanks, I'm just reading about it now.

Re CSS, I'm trying to keep hierarchical structure, but just make some containers to look like cells (proper dimensions and alignment). I'll read more about CSS grid and see if I can use it.

webpack.config.js
```diff
@@ -37,12 +37,28 @@ module.exports = {
...
-        test: /\.less$/,
-        loader: 'style!css!less!'
+        test: /css\/dashboard\/.*\.less$/,
+        loaders: [
+          'style-loader',
+          {
+            loader: 'css-loader',
+            options: {
+              modules: true,
+              sourceMap: true,
```
knowuh: I first saw this `modules` option when working with Portal Pages.  Mostly I think its cool, but the behavior wasn't what I was expecting.  And I think maybe worth discussion?

I find for example, I sometimes need to specify `:global(.my-selector) { … }` to intentionally override portal styles -- usually when existing selectors are tightly bound using ID values eg `:global(#left-nav) { … SCSS sub selectors here }`

The other thing which is can be tricky with modules option, is adding computed, or multiple `className`s to component using the module.  Like if I generate a list of class names using conditional string concatenation, and then want to add local scope class names, I have to do something like this:

```
    const linkStyles = (classNames) =>
      classNames.split(' ')
        .map((name) => css[name] || name)
        .join(' ')
        .replace(/^\s+|\s+$/g, '')
```

(That example might not make a lot of sense out of context. Its a special case because it uses both local classNames and classNames from the portal …)

I believe that another option is to use global name spacing by default, and then within SCSS specify local class names.  Using SCSS you could then wrap all your selectors in a local scope doing something like:

```
:local(.my-component-class) {
    h1 { 
       font-size: 10vw ; 
    }
}

```

Just thinking about this outloud because I haven't used css-loader modules much yet.

pjanik: I guess it's great to have this discussion, as it's probably mostly me advocating for CSS modules so far. 😉 

> I find for example, I sometimes need to specify :global(.my-selector) { … } to intentionally override portal styles -- usually when existing selectors are tightly bound using ID values eg :global(#left-nav) { … SCSS sub selectors here }

The main reason why I wanted to introduce CSS modules in Portal Pages is to be **not** able to overwrite Portal styles and to avoid conflicts with them. I think that was part of the Portal Pages dev guide that @scytacki shared and I agree with this idea. I've never actually used :global or :local modifiers. Can't we just move all the styles to Portal Pages and have self contained components, and then remove styles from Portal? Although I can imagine that navigation can be more tricky than a typical page.

> The other thing which is can be tricky with modules option, is adding computed, or multiple classNames to component using the module. Like if I generate a list of class names using conditional string concatenation, and then want to add local scope class names, I have to do something like this: (...)

If you don't need to use global / Portal class names, it's pretty much concatenation of string variables instead of string literals, so I'd say there's no difference. In this example I get the idea, but I'm not sure what replace does. So, it seems the problem is again about mixing local and global class names.

Could you describe this use case where you need Portal classes? I was naively assuming that we would never need that.

knowuh: In the portal pages case, I was trying to use the same div ID values for fear of breaking tests.

CSS selectors that specify the ids in them like `#navigation li { … }` are used in the portal.  

These ID based selectors have tighter binding (more specificity) than any local class-named styles -- and so I can't easily change them.  I suppose I could use `important` -- but that seems like a desperate measure.  

But maybe I should consider it more carefully.

knowuh: I just had a moment of clarity re-reading your comments.


You are correct, if we remove styles from the portal, this won't be a problem.  Perhaps this is where I should focus my efforts.

But still, I wonder about having the default be to localize everything.  I believe that by disabling `module: true` in the configuration, and using `:local(.component-class-name) { … <nested selectors> … }` in our sass files we will derive all the isolation benefits css modules at the component level with minimal impact and have fewer `css['style-name']` lookups.

… just thinking

knowuh: Also: I do think css modules are cool.  Thanks for pointing them out to me.

pjanik: > These ID based selectors have tighter binding (more specificity) than any local class-named styles -- and so I can't easily change them. I suppose I could use important -- but that seems like a desperate measure.

In this case, I'd just remove styles from Portal. Styling is spread between Portal Pages and Portal is something that I'd personally strongly avoid.

> But still, I wonder about having the default be to localize everything. I believe that by disabling module: true in the configuration, and using :local(.component-class-name) { … <nested selectors> … } in our sass files we will derive all the isolation benefits css modules at the component level with minimal impact and have fewer css['style-name'] lookups.

Perhaps it's a matter of taste and I guess we should discuss with other devs what would be the best practice. But I've worked on a couple of projects where 100% styles are localized and I can see only benefits. I think it's way harder to create mess using local styles. You're **forced** to have component-specific styles. You're sure that your elements are not modified by some more general styles. Also, my IDE does a great job of letting me jump directly to CSS class from JS file (but that's less important). For theming and sharing some properties we can use LESS/SASS variables and mixins, or some theming solutions based on CSS modules.

Honestly, I don't see a single use case where global styles are useful or more beneficial. It only lets you create a bad pattern of general, overlapping styles. And the situation that we have in LARA and Portal where I have no idea what's modified by what, and I just keep adding new layers of styles on top of the old CSS mess.

The only thing that comes to my mind is theming. Global styles theoretically allow you do it easier, but I think there's some elegant way to do it using CSS modules (e.g. react toolbox does it that way, we could look for some inspiration if we need to implement it).

Your solution with local would work, but it you need to explicitly enable localization which means that someone will forget to do it sooner or later.  That's why personally I'd rather enable modules by default and let someone use global style if necessary, rather than the other way. 

> have fewer css['style-name'] lookups.

By lookups you mean `<div className={css.myClass}>` vs `<div className='my-class'>`?
Personally, I'd say these two are totally equal. It's 4 characters more, and that's all. Also, when I concatenate classes, I just do:
`className = css.someClass + ' ' + css.someOtherClass`
instead of
`className = 'some-class' +  ' ' + 'some-other-class'`

But it's only my opion, based on my personal work style and patterns, so I guess it'd be great topic for some standup or dev meeting. I might be missing some use cases or patterns used by other devs that are useful too.

pjanik:
@knowuh, thanks for pointing out scrolling sync library. I'll keep that in mind and try to refactor it later.

PR #63, opened 2018-06-19
Add bunch of unit tests, mostly reducers and selectors

pjanik:
Still working on other tests, but this set seems pretty straightforward and finished, so opening a smaller PR.


test/selectors/dashboard-selectors_spec.js
```diff
@@ -0,0 +1,82 @@
...
+        },
+        dashboard: {
+          sortBy: 'name'
+        }
+      })
+
+      it('should return sorted list of students', () => {
+        expect(getSortedStudents(state).toJS()).to.eql(
+          // Students sorted by name (last name first)
+          [ s3, s2, s1 ]
```
ekosmin: I would add a test for first name sorting here. Looking at the selector, there's actually a bug where these students aren't sorted correctly:

```
const s1 = { id: 2, firstName: 'Y', lastName: 'AB' }
const s2 = { id: 3, firstName: 'Z', lastName: 'A' }
```

`s2` should come first, but `s1` comes first because the names are concatenated.

test/selectors/report-tree_spec.js
```diff
@@ -0,0 +1,131 @@
...
+    children: [ expectedActivityTrees({ questionVisible })[1] ]
+  })
+
+  describe('getAnswerTrees', () => {
+    it('should return answers with students ids mapped to students', () => {
+      expect(getAnswerTrees(state({})).toJS()).to.eql(expectedAnswerTrees)
+    })
+  })
+
+  describe('getQuestionTrees', () => {
```
ekosmin: Nice nested tests!

ekosmin:
Oh and one thing I should have mentioned in a previous PR: I'm having trouble getting `webpack-dev-server` to run on a fresh clone. I needed to drop to version 2 for it to work with version 3 of `webpack`

pjanik:
@ekosmin, really good catch about sorting, thanks! Fixed it and updated tests to check this use case.
When you run `webpack-dev-server`, you usually use a global one. That's why it was working for me. I've just added "server" script to package.json and specified correct version of webpack-dev-server. Now you can run `npm run server` to make sure that the correct one will be used.

PR #64, opened 2018-06-20
Sort students

ekosmin:
Hi @pjanik! I put together the dropdown to sort students today - thanks for leaving clues as to how it should work in the code :) Everything went pretty smoothly, including testing, but I do have a few questions I'll leave below. 

It's also worth noting I opened this request off your `tests` branch to pull in some helpful logic from your tests. This will probably cause some conflicts down the line, but I'm willing to resolve them as they come up.

### DIFF
--- a/js/actions/dashboard.js
+++ b/js/actions/dashboard.js
@@ -1,5 +1,10 @@
 export const SET_ACTIVITY_EXPANDED = 'SET_ACTIVITY_EXPANDED'
 export const SET_STUDENT_EXPANDED = 'SET_STUDENT_EXPANDED'
+export const SET_STUDENT_SORT = 'SET_STUDENT_SORT'
+
+export const SORT_BY_NAME = 'NAME'
+export const SORT_BY_MOST_PROGRESS = 'MOST_PROGRESS'
+export const SORT_BY_LEAST_PROGRESS = 'LEAST_PROGRESS'
 
 export function setActivityExpanded (activityId, value) {
   return (dispatch, getState) => {
@@ -37,5 +42,12 @@ export function setStudentExpanded (studentId, value) {
         dispatch(setActivityExpanded(firstActivity.get('id'), true))
       }
     }
+  }
+}
+
+export function setStudentSort (value) {
+  return {
+    type: SET_STUDENT_SORT,
+    value
   }
 }
--- a/js/components/dashboard/dropdown.js
+++ b/js/components/dashboard/dropdown.js
@@ -1,0 +1,24 @@
+import React, { PureComponent } from 'react'
+import { SORT_BY_NAME, SORT_BY_MOST_PROGRESS, SORT_BY_LEAST_PROGRESS } from '../../actions/dashboard'
+
+export default class Dropdown extends PureComponent {
+  constructor (props) {
+    super(props)
+    this.onSortSelected = this.onSortSelected.bind(this)
+  }
+
+  onSortSelected (e) {
+    const { setStudentSort } = this.props
+    setStudentSort(e.target.value)
+  }
+
+  render () {
+    return (
+      <select onChange={this.onSortSelected}>
+        <option value={SORT_BY_NAME}>Sort by Name</option>
+        <option value={SORT_BY_MOST_PROGRESS}>Sort by Most Progress</option>
+        <option value={SORT_BY_LEAST_PROGRESS}>Sort by Least Progress</option>
+      </select>
+    )
+  }
+}
--- a/js/reducers/dashboard-reducer.js
+++ b/js/reducers/dashboard-reducer.js
@@ -1,8 +1,8 @@
 import { Map } from 'immutable'
-import { SET_ACTIVITY_EXPANDED, SET_STUDENT_EXPANDED } from '../actions/dashboard'
+import { SET_ACTIVITY_EXPANDED, SET_STUDENT_EXPANDED, SET_STUDENT_SORT, SORT_BY_NAME } from '../actions/dashboard'
 
 const INITIAL_DASHBOARD_STATE = Map({
-  sortBy: 'name',
+  sortBy: SORT_BY_NAME,
   expandedActivities: Map(),
   expandedStudents: Map()
 })
@@ -13,6 +13,8 @@ export default function dashboardReducer (state = INITIAL_DASHBOARD_STATE, actio
       return state.setIn(['expandedActivities', action.activityId.toString()], action.value)
     case SET_STUDENT_EXPANDED:
       return state.setIn(['expandedStudents', action.studentId.toString()], action.value)
+    case SET_STUDENT_SORT:
+      return state.setIn(['sortBy'], action.value)
     default:
       return state
   }
--- a/js/reducers/dashboard-reducer.js
+++ b/js/reducers/dashboard-reducer.js
@@ -14,7 +14,7 @@ export default function dashboardReducer (state = INITIAL_DASHBOARD_STATE, actio
     case SET_STUDENT_EXPANDED:
       return state.setIn(['expandedStudents', action.studentId.toString()], action.value)
     case SET_STUDENT_SORT:
-      return state.setIn(['sortBy'], action.value)
+      return state.set('sortBy', action.value)
     default:
       return state
   }
--- a/js/selectors/dashboard-selectors.js
+++ b/js/selectors/dashboard-selectors.js
@@ -1,5 +1,6 @@
 import { createSelector } from 'reselect'
 import { getActivityTrees } from './report-tree'
+import { SORT_BY_NAME, SORT_BY_MOST_PROGRESS, SORT_BY_LEAST_PROGRESS } from '../actions/dashboard'
 
 // Inputs
 const getStudents = state => state.getIn(['report', 'students'])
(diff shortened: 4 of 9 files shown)


js/actions/dashboard.js
```diff
@@ -1,5 +1,10 @@
 export const SET_ACTIVITY_EXPANDED = 'SET_ACTIVITY_EXPANDED'
 export const SET_STUDENT_EXPANDED = 'SET_STUDENT_EXPANDED'
+export const SET_STUDENT_SORT = 'SET_STUDENT_SORT'
+
+export const SORT_BY_NAME = 'NAME'
+export const SORT_BY_MOST_PROGRESS = 'MOST_PROGRESS'
+export const SORT_BY_LEAST_PROGRESS = 'LEAST_PROGRESS'
```
ekosmin: I added the sort constants to the actions...but maybe they're better off elsewhere? I notice that there are imports like `import * as types from '../../js/actions/dashboard'` and adding non-type constants sort of messes that up.

js/components/dashboard/dropdown.js
```diff
@@ -0,0 +1,24 @@
+import React, { PureComponent } from 'react'
+import { SORT_BY_NAME, SORT_BY_MOST_PROGRESS, SORT_BY_LEAST_PROGRESS } from '../../actions/dashboard'
+
+export default class Dropdown extends PureComponent {
```
ekosmin: I called this `Dropdown` thinking that it could be made generic moving forward, even though it's only for sorting now. Do you think it's ok to leave like this?

js/selectors/dashboard-selectors.js
```diff
@@ -44,14 +45,52 @@ export const getStudentProgress = createSelector(
...
+
+// Returns sorted students
 export const getSortedStudents = createSelector(
-  [ getStudents, getDashboardSortBy ],
-  (students, sortBy) => {
-    if (sortBy === 'name') {
-      return students.toList().sortBy(student =>
-        (student.get('lastName') + student.get('firstName')).toLowerCase()
-      )
+  [ getStudents, getDashboardSortBy, getStudentTotalProgress ],
```
ekosmin: I call the `getStudentTotalProgress` selector here, even though it doesn't always get used when sorting by name. Will this cause unnecessary re-calculations?

test/selectors/dashboard-selectors_spec.js
```diff
@@ -60,23 +64,44 @@ describe('dashboard selectors', () => {
...
-      })
+      )
+    })
+  })
 
+  describe('getSortedStudents', () => {
+    describe('when sorting by name', () => {
       it('should return sorted list of students', () => {
-        expect(getSortedStudents(state).toJS()).to.eql(
+        expect(getSortedStudents(state({})).toJS()).to.eql(
```
ekosmin: Because the sorting selector now pulls in student progress, I needed to use the bigger state from above, which makes for a much messier test. Do you know if there's a way to stub out selectors so I can get away with passing a smaller state?

test/selectors/dashboard-selectors_spec.js
```diff
@@ -60,23 +64,44 @@ describe('dashboard selectors', () => {
...
           // Students sorted by name (last name first)
           [ s3, s2, s1 ]
         )
       })
     })
+
+    describe('when sorting by most progress', () => {
+      it('should return sorted list of students', () => {
+        expect(getSortedStudents(state({sortBy: SORT_BY_MOST_PROGRESS})).toJS()).to.eql(
```
ekosmin: The ability to stub out selectors would really help here. Since I already test `getStudentTotalProgress` above, it would be nice to just mock its output here. Right now, this test does too much.

js/selectors/dashboard-selectors.js
```diff
@@ -44,14 +45,52 @@ export const getStudentProgress = createSelector(
...
+  (students, studentProgress) => {
+    return students.map(student => {
+      const id = student.get('id').toString()
+      return studentProgress.get(id).toList().reduce((sum, progress) => sum + progress)
+    })
+  }
+)
+
+// A comparison function to sort students by last and then first name
+const compareStudentsByName = (student1, student2) => {
```
ekosmin: I made a comment on your PR about sorting - this is the solution I had in mind. I ended up wanting it to break ties for students with the same progress, let me know what you think

test/selectors/dashboard-selectors_spec.js
```diff
@@ -60,23 +64,44 @@ describe('dashboard selectors', () => {
...
-          students: { 1: s1, 2: s2, 3: s3 }
-        },
-        dashboard: {
-          sortBy: 'name'
+  describe('getStudentTotalProgress', () => {
+    it('should return hash with student total progress', () => {
+      expect(getStudentTotalProgress(state({})).toJS()).to.eql(
+        {
+          1: 1.5,
+          2: 1.5,
```
pjanik: I'd rather normalize total progress to value from [0, 1] range. If we ever need to render some general progress bars, it will be easier to have this kind of value. I guess it should be very easy, just divide this value by number of activities if I understand things correctly.

test/selectors/dashboard-selectors_spec.js
pjanik: That's a good question. I don't know yet. I don't think it's too bad here though, as we shared this state between all the sorting methods. But it might get useful in the future.

test/selectors/dashboard-selectors_spec.js
pjanik: If you really want, note that you could create named function that is then passed to `createSelector`. Something like this:

```javascript
const getSortedStudents = (students, totalProgress) => { .... }
const getSortedStudentsSelector = createSelector([getStudents, getTotalProgress], getSortedStudents)
```
Export both, and you could test only test `getSortedStudents`. Then it's super easy to provide any argument you want, without providin the whole state.

However, as long as the test is pretty concise, readable, and setup isn't too bad, I'd say it's an advantage that we test more than less. Generally, I'm not a fan of tiny unit tests that stub and mock everything and test some tiny logic. I find that it's easy to create tests that actually test themselves, not the application anymore. :wink: But there's a thin line here. I guess that in case of some selectors, this approach can be helpful. It's a bit overhelming that you need to create the whole app state. Maybe even here it would make sense, but since the state-crafting work has been done already, so I wouldn't change it.

test/selectors/dashboard-selectors_spec.js
```diff
@@ -1,54 +1,58 @@
...
 import { fromJS } from 'immutable'
-import { getSortedStudents, getStudentProgress } from '../../js/selectors/dashboard-selectors'
+import { getSortedStudents, getStudentProgress, getStudentTotalProgress } from '../../js/selectors/dashboard-selectors'
+import { SORT_BY_NAME, SORT_BY_MOST_PROGRESS, SORT_BY_LEAST_PROGRESS } from '../../js/actions/dashboard'
 
 describe('dashboard selectors', () => {
   const s1 = { id: 1, firstName: 'X', lastName: 'C' }
-  const s2 = { id: 2, firstName: 'y', lastName: 'b' }
+  const s2 = { id: 2, firstName: 'Y', lastName: 'AB' }
   const s3 = { id: 3, firstName: 'Z', lastName: 'A' }
```
pjanik: I've updated these names in my branch a bit to test your use case. Actually, I've used lower case on purpose to test if we ignore it. You can use them (or update these), as they test both lower vs uppercase and concatenation issue.

js/selectors/dashboard-selectors.js
pjanik: It was a good catch, thanks. You can check my fix, as it's a bit simpler. It's just enough to concatenate last name and first name with whitespace in between. JS will do all the work for us.

js/reducers/dashboard-reducer.js
```diff
@@ -13,6 +13,8 @@ export default function dashboardReducer (state = INITIAL_DASHBOARD_STATE, actio
       return state.setIn(['expandedActivities', action.activityId.toString()], action.value)
     case SET_STUDENT_EXPANDED:
       return state.setIn(['expandedStudents', action.studentId.toString()], action.value)
+    case SET_STUDENT_SORT:
+      return state.setIn(['sortBy'], action.value)
```
pjanik: I guess `setIn` isn't necessary here, `.set('sortBy', action.value)` would work just fine, but it's pretty much whatever.

js/components/dashboard/dropdown.js
pjanik: I'd rather call it SortByDropdown or something like this. Right now it isn't generic at all and this name can be a bit misleading. If we ever need to create a generic Dropdown, we'll just do it, I don't think it's a big deal.

js/actions/dashboard.js
pjanik: Seems fine for me since we need to share that between component, action, and tests. This seems like the best place.

js/selectors/dashboard-selectors.js
pjanik: Yes, but I wouldn't worry about it. It's just one edge case when we sort by name. It's not that expensive to sort students and student total progress won't change that often (even if it's every few seconds).

js/selectors/dashboard-selectors.js
pjanik: Kirk's comment:
> To sort names correctly in the general case, first names and last names should be compared independently. Either can contain spaces, for instance.

So, I'm wrong above. Your approach is better.

ekosmin: Got it! I did throw a `.toLocaleLowerCase()` in there so it handles capitalization like yours does

ekosmin:
@pjanik Thanks again for taking a close look through my code and answering my questions! I agree on your points about the tests - if the state becomes a problem moving forward I'll reconsider.

PR #66, opened 2018-06-21
Replacing dummy answer cell with a generic answer icon

ekosmin:
A very simple component that shows an icon when there's any answer. This is the default cell now, but I renamed the existing `AnswerCell` to `ShortAnswer`, since it should be a useful basis for that in the future.

http://portal-report.concord.org/branch/generic-icons/index.html?dashboard=true

### DIFF
--- a/css/dashboard/helpers.less
+++ b/css/dashboard/helpers.less
@@ -24,6 +24,7 @@
   line-height: @row-height;
   padding: @row-padding;
   border: @cell-border;
+  flex-grow: 1;
 }
 
 .nowrap {
--- a/js/components/dashboard/generic-answer.js
+++ b/js/components/dashboard/generic-answer.js
@@ -1,0 +1,17 @@
+import React, { PureComponent } from 'react'
+
+import css from '../../../css/dashboard/generic-answer.less'
+
+export default class GenericAnswerCell extends PureComponent {
+  render () {
+    const { answer } = this.props
+    return (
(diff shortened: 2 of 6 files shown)


css/dashboard/helpers.less
```diff
@@ -24,6 +24,7 @@
   line-height: @row-height;
   padding: @row-padding;
   border: @cell-border;
+  flex-grow: 1;
```
ekosmin: I moved this into the basic `cell` css...I think it is a more reasonable default than `0`, since in general the cells want to take up the whole container.

js/components/dashboard/generic-answer.js
```diff
@@ -0,0 +1,17 @@
+import React, { PureComponent } from 'react'
+
+import css from '../../../css/dashboard/generic-answer.less'
+
+export default class GenericAnswerCell extends PureComponent {
+  render () {
+    const { answer } = this.props
```
ekosmin: I didn't end up needing an `expanded` prop since rendering the icon was so simple, unless I'm missing something.

pjanik:
Looks good, thanks!

pjanik:
OK, for the future reference - we will need to check if the answer is submitted and if the type is !== "NoAnswer". Currently, that renders an icon even if the student hasn't provided an answer or hasn't submitted it. I'll probably fix it while working on other question types.

PR #69, opened 2018-06-22
Use offering URL as fallback when reportUrl is not available

pjanik:
[#158501911]

@scytacki, I've implemented it here, as that's what you seemed to prefer.
Note that we'll need to change Portal soon anyway to support highlighted questions, so we could consider adding `reportUrl` to external reports too. I think we don't do it only for "historical" reasons. It won't have any performance consequences until report actually uses this URL (that it can generate anyway, as that's what we do here). 

Also, when I add `http://localhost:8080/?dashboard=true` as an external report, the final URL is `http://localhost:8080/?dashboard=true?offering=...`. There are two question marks. I think you pointed out this issue recently. That's another reason to modify this Portal code. However, it does not break anything, the browser seems to handle it correctly anyway.

Anyway, this PR ensures that we can start using Dashboard right away.

### DIFF
--- a/js/api.js
+++ b/js/api.js
@@ -7,17 +7,32 @@ const warn = (message) => {
     console.warn(message)
   }
 }
-const urlParams = (() => {
-  return queryString.parse(window.location.search)
-})()
+const urlParam = (name) => {
+  return queryString.parse(window.location.search)[name]
+}
 
 // Report URL and auth tokens are provided as an URL parameters.
-const REPORT_URL = urlParams['reportUrl']
-const AUTH_HEADER = `Bearer ${urlParams['token']}`
+const getReportUrl = () => {
+  const reportUrl = urlParam('reportUrl')
+  const offeringUrl = urlParam('offering')
+  if (reportUrl) {
+    return reportUrl
+  }
+  if (offeringUrl) {
+    // When this report is used an external report, it will be launched with offering URL parameter instead of reportUrl.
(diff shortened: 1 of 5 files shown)


js/api.js
```diff
@@ -7,17 +7,32 @@ const warn = (message) => {
     console.warn(message)
   }
 }
-const urlParams = (() => {
-  return queryString.parse(window.location.search)
-})()
+const urlParam = (name) => {
+  return queryString.parse(window.location.search)[name]
+}
```
kswenson: With this change the URL params are parsed on each call rather than once and cached. In some apps we remove URL params like short-lived tokens during the running of the app. If that's a possibility here then parsing the URL once and caching the results may be preferable.

pjanik: Good point, thanks. Although I've changed it so this module can be tested (tests are also part of this PR). If it's cached in the module level (like it used to be before), once you import a module in your test suite, you can't do anything. At this point it doesn't seem likely we're going to hide any of these params, so I'd rather keep it testable for now. Not sure if there's some easy solution that would handle both cases well.

kswenson: What I've done in several projects is have `urlParams` in its own module, so that it can be imported/required by any module that needs it. For testing purposes it could then be mocked, but that may be more effort than it's worth at this point.

pjanik:
@ekosmin, could you take a look when you get a chance? I guess Scott's PR queue is long, and it'd be useful for me to merge it asap.

PR #70, opened 2018-06-26
Open Response Answer + Cypress Tests

ekosmin:
Hi @pjanik! This PR has two parts: the open response answer cell, and Cypress tests. I had a little bit of trouble getting it to run on Travis, though it's mostly working now. 

Let me know if you'd rather pass off some of the Cypress review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ language:
   - ruby
 node_js:
   - node
+cache:
+  directories:
+    - ~/.npm
+    - ~/.cache
 install:
   - travis_retry gem install s3_website -v 3.4.0
   - travis_retry npm install
@@ -11,7 +15,11 @@ before_script:
   - webpack
   - WEBPACK_TARGET=lib webpack
 script:
-  - "npm run test"
+  - npm run test
+  - npm run server &
+  - cypress install
+  - wait-on http://localhost:8080
+  - $(npm bin)/cypress run --config video=false
   - ./s3_deploy.sh
 cache:
   bundler: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,14 +10,13 @@ cache:
     - ~/.cache
 install:
   - travis_retry gem install s3_website -v 3.4.0
-  - travis_retry npm install
+  - travis_retry npm ci
 before_script:
   - webpack
   - WEBPACK_TARGET=lib webpack
 script:
   - npm run test
   - npm run server &
-  - cypress install
   - wait-on http://localhost:8080
   - $(npm bin)/cypress run --config video=false
   - ./s3_deploy.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,17 +4,12 @@ language:
   - ruby
 node_js:
   - node
-cache:
-  directories:
-    - ~/.npm
-    - ~/.cache
 install:
   - travis_retry gem install s3_website -v 3.4.0
   - travis_retry npm ci
-before_script:
+script:
   - webpack
   - WEBPACK_TARGET=lib webpack
-script:
   - npm run test
   - npm run server &
   - wait-on http://localhost:8080
@@ -24,3 +19,5 @@ cache:
   bundler: true
   directories:
     - node_modules
+    - ~/.npm
+    - ~/.cache
--- a/cypress/integration/dashboard.spec.js
+++ b/cypress/integration/dashboard.spec.js
@@ -1,0 +1,34 @@
+import React from 'react'
+
+describe('Dashboard', function () {
+  beforeEach(() => {
+    cy.visit('http://localhost:8080/?dashboard=true')
+  })
+
+  let rowHeight = null
+  it ('Has equal height rows', function () {
+    cy.get('[class*=studentName-]').should('be.visible').then((students) => {
+      Array.from(students).forEach((student) => {
+        if (rowHeight == null) {
+          rowHeight = student.clientHeight
+        }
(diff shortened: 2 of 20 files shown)


package.json
```diff
@@ -24,8 +24,11 @@
   "scripts": {
     "server": "webpack-dev-server",
     "prepare": "WEBPACK_TARGET=lib webpack",
+    "cypress:test": "cypress run --config video=false,trashAssetsBeforeRuns=false",
+    "cypress:open": "cypress open",
     "test": "mocha --compilers js:babel-core/register --require ignore-styles --require ./test/test_helper.js 'test/**/*.@(js|jsx)'",
-    "test:watch": "npm run test -- --watch"
+    "test:watch": "npm run cypress:open & npm run test -- --watch",
+    "test:full": "npm run test && npm run cypress:test"
```
ekosmin: I wanted to get all the tests running through Cypress, but I couldn't make it happen...so I added a handful of new testing commands instead. I left `npm run test` alone since the Cypress tests take a while to start up, but they're fine while `watch`ing so I added them there.

js/components/dashboard/answer.js
```diff
@@ -8,8 +8,10 @@ import css from '../../../css/dashboard/answer.less'
 export const NoAnswer = () => null
 
 export const GenericAnswer = () => (
-  <div style={{textAlign: 'center'}}>
-    <i className='icomoon-radio-unchecked' />
+  <div className={css.genericAnswer}>
+    {
+      '\u2b24' // large circle
```
ekosmin: Trudi requested we show a filled-in circle instead...I just went with the unicode one.

cypress/utils.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+
+  // CSS modules randomize class names, so it is safest to select with explicit data-* attributes
+  // https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements
+  getByCypressTag: function ( cypressTag ) {
+    return cy.get('[data-cy=' + cypressTag + ']')
```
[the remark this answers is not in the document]
kswenson: This could also be written as:
```javascript
  return cy.get(`[data-cy=${cypressTag}]`)
```

.travis.yml
```diff
@@ -4,14 +4,22 @@ language:
   - ruby
 node_js:
   - 6
+cache:
+  directories:
+    - ~/.npm
+    - ~/.cache
```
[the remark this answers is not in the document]
pjanik: This guide seems complete. You can switch Node.js version that we're using a few lines above. Try:
```
node_js:
  - node
```
instead of `- 6`. It should install the latest stable Node.js (and NPM), perhaps v10. Then you should have access to `npm ci`.

cypress/utils.js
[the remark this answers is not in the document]
pjanik: Nice, a great guide that answers some of the yesterday's questions. I think it's a better approach than relying on CSS classes used for styling that can change anytime.

package.json
pjanik: I'd definitely keep unit tests separate from Cypress. There's no reason to use Cypress machinery for that. It's more likely that developer is already familiar with Mocha, no startup time, it's faster, and Mocha won't die, while Cypress might (I hope not). I'd say it's still very easy to run Cypress, but it's pretty opinionated tool.

cypress/utils.js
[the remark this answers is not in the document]
ekosmin: Thanks @kswenson! I should get to know template literals better...

package.json
ekosmin: Agreed that I wouldn't want to depend too much on Cypress! The only reason I had the thought is because Cypress just bundles most of the tools we use by default (Mocha, Chai, and Sinon), so it should be easy to move back and forth. Still, I see what you mean - I'll leave them separate for now.

https://docs.cypress.io/guides/references/bundled-tools.html#

.travis.yml
```diff
@@ -3,15 +3,22 @@ language:
...
 install:
   - travis_retry gem install s3_website -v 3.4.0
-  - travis_retry npm install
+  - travis_retry npm ci
 before_script:
   - webpack
   - WEBPACK_TARGET=lib webpack
```
kswenson: Unrelated to this PR, but in reading up on Travis earlier today I came across a treatise on why build steps should be part of `script` rather than `before_script`. Unfortunately, I don't have the link handy, but the essence of it was that failures in `before_script` are considered the same as failures in `install`, i.e. they are reported like external failures that could potentially be resolved by simply trying again because maybe there was a network hiccup during the install process. Failures in the `script` phase are the only ones the are treated as build failures, but that's exactly how a webpack failure should be treated.

cypress/integration/dashboard.spec.js
```diff
@@ -6,6 +6,26 @@ describe('Dashboard', function () {
     cy.visit('http://localhost:8080/?dashboard=true')
   })
 
+  it ('Renders cells in a row with the same height', function () {
```
ekosmin: @pjanik Thanks for looking through this! I think I addressed all your comments, except for updating `cypress:test` to run the server, which is a good idea. I wanted to tag you on the new grid tests I added based on your suggestions - I learned about some new Cypress functions that make chaining much cleaner! They're all documented here:

https://docs.cypress.io/api/introduction/api.html#

pjanik:
Thanks @ekosmin! I went ahead and resolved some conflicts with master. Also, moved webpack build to script section following @kswenson advice (thanks for looking at our PRs and good comments!).

PR #71, opened 2018-06-26
Featured questions

pjanik:
This PR adds support of the featured questions. 

There's some small refactoring of how question visibility is calculated. I moved that to the tree selector, and I had to add the notion of questions which are "hiddenByUser" since the visibility is now more complex (depends on a few properties now - "isFeatured" and "hiddenByUser").

I hope the other bits should be clear. There's bunch of tests too. @ekosmin, there's example of testing selectors without providing the whole state. We can use `resultFunc` to test function directly and pass it some mock input.

### DIFF
--- a/js/actions/index.js
+++ b/js/actions/index.js
@@ -6,8 +6,8 @@ export const RECEIVE_ERROR = 'RECEIVE_ERROR'
 export const SET_NOW_SHOWING = 'SET_NOW_SHOWING'
 export const SET_ANONYMOUS = 'SET_ANONYMOUS'
 export const SET_QUESTION_SELECTED = 'SET_QUESTION_SELECTED'
-export const SHOW_SELECTED_QUESTIONS = 'SHOW_SELECTED_QUESTIONS'
-export const SHOW_ALL_QUESTIONS = 'SHOW_ALL_QUESTIONS'
+export const HIDE_UNSELECTED_QUESTIONS = 'HIDE_UNSELECTED_QUESTIONS'
+export const SHOW_UNSELECTED_QUESTIONS = 'SHOW_UNSELECTED_QUESTIONS'
 export const SET_ANSWER_SELECTED_FOR_COMPARE = 'SET_ANSWER_SELECTED_FOR_COMPARE'
 export const SHOW_COMPARE_VIEW = 'SHOW_COMPARE_VIEW'
 export const HIDE_COMPARE_VIEW = 'HIDE_COMPARE_VIEW'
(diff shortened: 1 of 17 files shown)


pjanik:
> When is showFeaturedQuestionsOnly set to false? I would maybe expect it to be false when the "Show All" button is pressed on the main report, but it doesn't look like that's the case. Reading the PT story, I can't tell if there is supposed to be a way to view non-featured questions on either report.

There's no way to set it to false yet. It will be possible in the future, but that's not the priority yet and it requires some discussion. For now, the assumption was that if some questions are defined as featured, only these questions should be displayed. Adding a toggle in Dashboard will be trivial. It's more complicated in normal report where we have some other options to hide questions.

> Is visible a redundant piece of state right now? It seems like it can be calculated from showFeaturedQuestionsOnly, isFeatured and isHidden.

You're right. But note that `visible` is **not** part of the state (aka redux store). It's only calculated by the selector. Just to make things simpler for child components. You can think of it as a "computed property" in MobX. That's what selectors are in fact.

> It seems like it should be sufficient to track isFeatured, selected and visible for each question. My thinking is that you could use isFeatured to initialize questions to selected and visible. A "Show All" button would set everything to selected and visible, and a "Show Featured" button would reset to a default state.

I don't really get this part. But as we agreed, 'visible' is just a computed property, not state. `isFeatured` has nothing to do with selection. I think we don't really want to mix these things. Selection is used to custom hide-show feature in the old report. It's also saved in teacher's profile.

I was considering whether "Show All" button should also show non-featured questions too. It would definitely make sense in the future. But for now, featured questions filtering is not directly visible to teachers. So, I decided that right now it shouldn't. We're missing some other controls at the moment. If you opened report with featured questions, clicked "Show All" and saw non-featured questions, there'd be no way to go back to featured-only view (initial one). We'll need some new controls, something like "Show featured only". But we would need to discuss that with Dan first. And I guess there's way to make all that buttons a bit simpler. 

So, the old report respects "isFeatured" too, but in a transparent way. It just limits display to these questions. Note that pretty much all the of the users of the old report won't use "isFeatured" for now, so that seems to be okay. I didn't want to intruduce new buttons there and confuse them.

I hope it makes things easier to understand, let me know if something is still unclear.

ekosmin:
Thanks for your response @pjanik! Really good point about `visible` being like a computed property.

I see that the functionality could expand in the future, but it is still unclear to me why `hiddenByUser` needed to be introduced to the state, instead of just using `visible` as it did before. If we have a chance, it would be useful for me to talk through a situation where just using `visible` is insufficient, or more difficult.

But I do think this all looks fine for now!

pjanik:
Previously:
`visible` = `!hiddenByUser`, one property was enough (no matter what we called that) and no need for computed state
Now:
`visible` = `!hiddenByUser && isFeatured`, two properties necessary, we need to calculate visibility using them
 
So, what I actually did was to rename `visible` to `hiddenByUser`. As that's explicit state and related action. And I had to intruduce computed property called `visible` that depends both on `hiddenByUser` and `isFeatured`.

PR #72, opened 2018-06-28
Adding a button to show/hide all students

ekosmin:
@pjanik, I ended up taking a first pass at the show/hide all students button:

http://portal-report.concord.org/branch/expand-all/index.html?dashboard=true

I put all the functionality into a single button rather than two - it closes the students if any are open. The behavior feels intuitive to me, let me know what you think. Also, this is the first PR with a *FULL SUITE* of tests, from unit to integration and E2E.


css/dashboard/expand-students.less
```diff
@@ -0,0 +1,6 @@
+.expandStudents {
+  width: 120px;
+  position: absolute;
+  left: 30px;
+  top: 200px;
```
ekosmin: Ideally the button would go in a cell next to the question row which expands/shrinks along with the questions. But, I just put it in a corner for now, since I wasn't sure if we would end up wanting 2 buttons anyway.

js/actions/dashboard.js
```diff
@@ -36,11 +45,20 @@ export function setStudentExpanded (studentId, value) {
...
-        dispatch(setActivityExpanded(firstActivity.get('id'), true))
-      }
+      checkActivityExpanded(dispatch, getState)
+    }
+  }
+}
+
+export function setStudentsExpanded (studentIds, value) {
+  return (dispatch, getState) => {
+    dispatch({
```
ekosmin: I wasn't sure the best way to dispatch actions to expand/close all the students. I could have just issued `SET_STUDENT_EXPANDED` actions for all students, but I figured that would make the action log unnecessarily confusing. I also considered making a `SET_ALL_STUDENTS_EXPANDED` action, but at that point I need to pass all the student IDs anyway since they're not in the Dashboard's state. This option seemed the most flexible.

js/components/dashboard/expand-students.js
```diff
@@ -0,0 +1,28 @@
...
+  onSetStudentsExpanded (value) {
+    const { setStudentsExpanded, students } = this.props
+    setStudentsExpanded(students.map(student => student.get('id')), value)
+  }
+
+  render () {
+    const { expandedStudents } = this.props
+    const anyStudentsExpanded = expandedStudents.some((isExpanded) => isExpanded)
+    return (
+      <Button onClick={() => this.onSetStudentsExpanded(!anyStudentsExpanded)} className={css.expandStudents}>
```
ekosmin: I know we talked about not passing arrow functions to React components...but it looks like the `Button` only re-renders when clicked, and the behavior is also here:

https://github.com/concord-consortium/portal-report/blob/master/js/components/report/report.js#L84

js/actions/dashboard.js
pjanik: You're not limited to Dashboard state only. Dashboard reducer is just reducer specific to Dashboard. The app itself uses bunch of common state (report itself). You actually have access to students here:
```javascript
const students = getState().getIn(['report', 'students'])
```
But what you did is okay too.

js/components/dashboard/expand-students.js
pjanik: You're definitely right, it doesn't matter here, Button doesn't have any children and expensive re-rendering. But I'd avoid it just to have a consistent style. And avoid a situation where someone copies that pattern in place where it actually matters. I know you pointed out some existing inconsistency, most likely added by me. :wink: But still, I wouldn't add new one.

If you don't like adding constructor just for that, you could also use method trick proposed by Kirk:
```
onClick = () => {
  ...
}
```
(note there's arrow function)
that provides auto-binding too. I'd vote for constructor as that's the main style in this project. Honestly, I never care about particular style (like here), but I do care more about consistency and common style. Most of CC project don't have that and it's painful for me. :wink:

js/components/dashboard/expand-students.js
```diff
@@ -0,0 +1,28 @@
+import React, { PureComponent } from 'react'
+import { Map, List } from 'immutable'
+import { SET_STUDENTS_EXPANDED } from '../../actions/dashboard'
+import Button from '../report/button'
```
pjanik: If some component gets reused, I'd move it to `../common/...`, so developer knows that he should check both app types while working on it (e.g. changing styles).

js/actions/dashboard.js
ekosmin: D'oh, right, thanks for the reminder!

js/components/dashboard/expand-students.js
ekosmin: Alright, I took care of this! I originally tried @kswenson's arrow function, but I realized I couldn't get it to work until I added the proper Babel transformer. This still didn't solve the problem (also in the code I referenced) of wanting to pass different parameters to the click handler in different cases. Binding them in the render function is just as bad as before, but I found this neat recommendation!

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md#protips

I ended up creating a sub-component for the button, so that I can pass different values to the click handler without binding functions in `props`.

ekosmin: As for a lack of consistency across CC projects...I don't want you to be in pain, but I think I do like the arrow functions over binding in the constructor! So I used them in the updated sub-component 😬

js/components/dashboard/expand-students.js
ekosmin: Done!

ekosmin:
Thanks again for the review @pjanik! I responded to all your comments, including updating the style to match Ethan's mock more closely.

cypress/integration/sort-by-dropdown.spec.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import { getByCypressTag } from '../utils'
+
+describe('Sort By dropdown', function () {
+  beforeEach(() => {
+    cy.visit('http://localhost:8080/?dashboard=true')
+  })
+
+  it ('Is visible', function () {
+    getByCypressTag('sortDropdown').should('be.visible')
```
ekosmin: The previous test was checking that the `Dashboard` was rendering a sorting dropdown, which I think this tests just fine. I briefly tried making an integration test for `DashboardApp` to see if it rendered the sorting dropdown, but ran into issues mounting a connected component.

ekosmin: Also I used to get an email whenever a build failed, but I didn't get any today? I'll have to check the configuration.

test/components/dashboard/expand-students_spec.js
```diff
@@ -1,25 +1,25 @@
...
+import { mount } from 'enzyme'
 import sinon from 'sinon'
 import { fromJS } from 'immutable'
 import ExpandStudents from '../../../js/components/dashboard/expand-students'
 
 describe('<ExpandStudents />', () => {
   it('should be labelled correctly on load', () => {
-    const wrapper = shallow(<ExpandStudents />)
-    assert(wrapper.find('Button').contains('Open Students'))
+    const wrapper = mount(<ExpandStudents />)
```
ekosmin: The only downside of the new sub-component approach is that it means I need to full `mount` the parent in testing instead of a `shallow` mount.

pjanik:
Looks good, thanks!

test/components/dashboard/expand-students_spec.js
pjanik: You could export `ExpandStudentsButton` too and test it separately using `shallow` (since it's unit testing). Perhaps it's never too many tests, although I think this is getting pretty far (testing specific labels).

You could also use `render`. I don't think you need full `mount` here.
Also, I'm not sure why there's division between `ExpandStudentsButton` and `ExpandStudents`. Not saying it's wrong, it's absolutely fine, but I think the reason to divide component would be when they get too big or e.g. to make testing easier. It's all good, just replying to your comment.

ekosmin: Thanks @pjanik! I'll try it with `render` instead, I'm still trying to figure out all the differences...

About the new component - the idea was to have an `onClick` function I could call with different parameters without binding them in `props`. I made a comment about it above but it might have gotten lost, I was following this suggestion I found from `eslint`:

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md#protips

Could be a useful pattern moving forward!

pjanik: FYI, they're mixing two very separate things, and take advantage of non-existing, old React API (`createReactClass`).

I agree with the first part - using `onClick={someFunction.bind(...)}` is bad. This is exactly the same as using arrow expresion  `onClick={() => { ... }}` which I keep constantly pointing out. And it's bad exactly due to the same reason. It creates a new function every time.

Adding a child component in this particular case also makes sense, as otherwise you would need to create some list of bound functions in the constructor - weird.

But then the only reason that they don't need to bind handler in child component is because `createReactClass` automatically binds all the methods. Note that you had to bind `onClick` in your `ExpandStudentsButton` class anyway. And you could have done the same thing in the parent component. I'd say there's no difference in this case.

But all in all, we're talking about really small things. ;) Althoug having some deeper understanding how React works is useful for sure.

PR #74, opened 2018-07-02
Rename is_featured to show_in_featured_question_report + scope question visibility filtering

pjanik:
[#158190868]

A few changes:
- is_featured -> show_in_featured_question_report
- featured question filtering is applied only in Dashboard
- user selection filtering is applied only in Full Report
- I've added the current view type to the reducer to implement things above, but it also means it'll be trivial to swap views in the future 
- I had to remove one test (pretty obvious logic), as now we read URL param on module load. URL reconfiguring won't have an effect as the module is already loaded in tests. I didn't have time to figure out how to work around that, and this logic is super simple anyway.


pjanik:
I need to setup test deployment for Trudi, so I'm going to merge it now. I don't think there'd be time for significant changes here, but I'm happy to address some smaller comments later if you have them @ekosmin.

PR #75, opened 2018-07-05
Add scored multiple choice questions summary

pjanik:
[#158441023]

This PR adds a new column with scored multiple choice questions summary. Trudi requested that. It will show up only if given activity has any scored multiple choice question.


pjanik:
I need to merge it pretty much right away, as it's last GEODE day and Trudi wanted to have that during upcoming tests. But if you get a chance @ekosmin, feel free to comment and I'll try to address that later.

PR #77, opened 2018-07-30
Update Dependencies (and related chores)

knowuh:
I am starting to work on this again, and in preparation I updated some old npm dependencies.

There were some cascading test failures. 

Tests on Dashboard were using shallow enzyme rendering.  

When enzyme was updated, ref methods of child nodes were no longer being invoked, and not setting instance variables on the Dashboard. So I had to switch to `enzyme.mount()` for those tests.

There were some other minor tweeks to silence some standardJS warnings.


package.json
```diff
@@ -32,40 +32,42 @@
     "test:full": "npm run test && npm run cypress:test"
   },
   "devDependencies": {
+    "autoprefixer": "^9.0.1",
```
pjanik: It seems it isn't used anywhere, right?

knowuh: Well, depends on what we mean by 'used'.  I am not certain we are using the functionality, so maybe we can remove it.  But we are using it as a plugin, so …

See https://github.com/concord-consortium/portal-report/blob/master/postcss.config.js

```
module.exports = {
  plugins: [
    require('autoprefixer')({ /* ...options */ })
  ]
}

```

I would be up for removing this completely, but I was trying to just clean up the deps enough to get started on work.

knowuh:
@pjanik regarding your questions about `ref`s and enzymes `shallow` render method, I am not sure why the tests were working before. IFIK this isn't a breaking change, its something that was *never* supported by shallow.  See this discussion from 2017: https://github.com/airbnb/enzyme/issues/1394

> ljharb: refs only work with mount, yes.

Here is the code where Dashboard uses refs:

We set `this.activityHeaders` using refs:  
`<div ref={el => { this.activityHeaders = el }} className={css.activityHeaders}>`

And then we expect `this.activityHeaders` to not be null when we are adding event handlers. (but this ref method won't get called by `shallow` rendering):

```
  componentDidMount () {
    // Make sure that the verticalScrollContainer fits the window height.
    window.addEventListener('resize', this.onResize)
    // Synchronize scrolling of headers and horizontalScrollContainer.
    this.horizontalScrollingContainer.addEventListener('scroll', this.onHorizontalContainerScroll)
    this.activityHeaders.addEventListener('scroll', this.onHeadersScroll)
    this.onResize()
  }
```

See:
https://github.com/concord-consortium/portal-report/blob/master/js/components/dashboard/dashboard.js#L98

package.json
pjanik: Thanks, I've looked only at webpack config and missed this one.

pjanik:
Okay, so `componentDidMount` is failing, that makes sense. Thanks for the explanation!

PR #78, opened 2018-08-08
159280307 expand question column

knowuh:
This PR allows report viewers to click on question headers (which are now displayed as `Q1.` … `Qn.` when activities are expanded) to open a vertical expansion of that question for all students.

The action and reducer were minimal changes.  Keeping the table cells well aligned involved some changes to CSS and to parent rendering methods. To simplify CSS I switched to `box-sizing: border-box` to eliminate a few computations that to into account padding values.

There will be some discussion about whether this feature makes sense, because as pointed out, all student row heights must be increased to view the question details.

This PR relates to two PT stories:
> Expand a single column on click so that the teacher can see full student answers for all students to that question. [#159280307](https://www.pivotaltracker.com/story/show/159280307)

and 
> Narrow columns to just display Q# and no question starter. [#159280335](https://www.pivotaltracker.com/story/show/159280335)

### DIFF
--- a/css/dashboard/helpers.less
+++ b/css/dashboard/helpers.less
@@ -7,11 +7,14 @@
 @student-name-width: 150px;
 @border-width: 1px;
 @cell-border: @border-width solid @main-dark;
-@row-height: 25px;
+@row-height: 40px;
 @row-padding: 8px;
 @row-outer-height: @row-height + 2 * @row-padding + 2 * @border-width;
 @expanded-student-row-height: 200px;
 @expanded-question-prompt-max-height: 100px;
+@expanded-question-width: 350px;
+@closed-question-width: 50px;
+
 
 // Transitions
 @transition-time: 1s;
@@ -21,13 +24,18 @@
 // Mixins
 .cell {
   min-height: @row-height;
-  line-height: @row-height;
   padding: @row-padding;
   border: @cell-border;
+  box-sizing: border-box;
 }
 
 .nowrap {
   white-space: nowrap;
   overflow: hidden;
   text-overflow: ellipsis;
+}
+
+.clickable {
+  cursor: pointer;
+  font-weight: 500;
 }
--- a/js/actions/dashboard.js
+++ b/js/actions/dashboard.js
@@ -7,6 +7,8 @@ export const SORT_BY_NAME = 'NAME'
 export const SORT_BY_MOST_PROGRESS = 'MOST_PROGRESS'
 export const SORT_BY_LEAST_PROGRESS = 'LEAST_PROGRESS'
 
+export const SET_QUESTION_EXPANDED = 'SET_QUESTION_EXPANDED'
+
 export function setActivityExpanded (activityId, value) {
   return (dispatch, getState) => {
     dispatch({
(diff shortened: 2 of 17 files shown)


css/dashboard/helpers.less
```diff
@@ -7,11 +7,14 @@
...
 @border-width: 1px;
 @cell-border: @border-width solid @main-dark;
-@row-height: 25px;
+@row-height: 40px;
 @row-padding: 8px;
 @row-outer-height: @row-height + 2 * @row-padding + 2 * @border-width;
 @expanded-student-row-height: 200px;
 @expanded-question-prompt-max-height: 100px;
+@expanded-question-width: 350px;
+@closed-question-width: 50px;
```
pjanik: These values seem to be defined both in CSS and JS now. I was trying to avoid that. 
I think that collapsed and expanded could be done by defining `flex: 1` (collapsed) and `flex: 100` (expanded). It should also help with cases when there are not many questions in a given activity. Then we wouldn't have to have these dimensions here too.
Otherwise, I've seen webpack LESS variables loader which lets you load values from this feel and have them available in JS. Although, it seems pretty exotic, so I'd rather try to avoid it.

js/actions/dashboard.js
```diff
@@ -27,6 +29,17 @@ export function setActivityExpanded (activityId, value) {
   }
 }
 
+export function setQuestionExpanded (questionId, value) {
+  return (dispatch, getState) => {
+    const expanded = getState().getIn(['dashboard', 'expandedQuestions', questionId.toString()])
+    dispatch({
+      type: SET_QUESTION_EXPANDED,
+      questionId,
+      value: !expanded
```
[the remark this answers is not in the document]
knowuh: Right -- good catch.  Thank you.

css/dashboard/helpers.less
knowuh: I was reluctant to do this, but I wasn't able to align columns correctly otherwise. I may try one last attempt with flex again or have a look at the variable loader you mention.

js/actions/dashboard.js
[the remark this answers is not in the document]
knowuh: Fix in b1f984f

css/dashboard/helpers.less
knowuh: I think I addressed these issues in in 62e7493

knowuh:
@pjanik  -- if you are willing to have another look, I added to commits to address your concerns:

* b1f984f -- Addresses `setQuestionExpanded` issue.
* 62e7493 -- Addresses hard-coded css widths for question prompt sizes.

knowuh:
Thanks @pjanik  -- You are right, those transitions were disorienting, and i didn't really notice it, or give them much thought.

I was able to follow your suggestions in e260df2

I also chained the vertical and horizontal animations, which solves the jumping question prompt height issue (#2) you brought up above.  I think it is also somewhat less disorienting then expanding both at the same time.  Curious for you to try it out.

pjanik:
Thanks, it works really cool now.
Re chaining, I think I like it. The only small issue is that when you have e.g. two students expanded, and you collapse one of them, there's only vertical transition, which is now delayed. So it seems like nothing is happening for a moment and then the animation is triggered. Not saying it's a big problem. I'm fine with either solution (chaining or not). 👍

PR #79, opened 2018-08-16
Show full question details when clicking on expansion icon in question header.

knowuh:
When a question prompt is expanded in the dashboard header, it also displays an expansion icon.

When the expansion icon is clicked, it opens up a question detail modal view, which displays the complete question prompt.  

This view component `components/dashboard/question-details.js` reuses components from the report question views.  These components were modified slightly so that properties configure optionally hidden buttons for providing feedback and comparing students. The default properties added to these reused components ensure behavior consistent for the old report if not specified.

One of the question prompts in the sample data was modified to include an image tag, so that QA could see what this looks like.

There are also some commits in this PR for a smaller story about adding Activity numbers to the table headers eg:  *Act 1: Act One Name* (where as before it would have been *Act One Name*)

This PR is for these two PT stories:

[Click question expand icon to pop up full question](https://www.pivotaltracker.com/story/show/159280381)

and 
[Add Act 1: Name of activity when there is a multi-activity sequence](https://www.pivotaltracker.com/story/show/159007047)


knowuh:
Sorry I let this PR get too large @pjanik   :/

js/components/dashboard/activity-questions.js
```diff
@@ -19,6 +21,10 @@ export default class ActivityQuestions extends PureComponent {
             expanded && activity.get('questions').filter(q => q.get('visible')).map(q => {
               const questionIsExpaned = expandedQuestions.get(q.get('id').toString())
               const expanded = showFullPrompts || questionIsExpaned
+              const openQuestionDetails = (e) => {
+                e.stopPropagation()
```
pjanik: Is that to prevent question collapsing? If so, some comment could be nice so someone doesn't remove it by accident (and tests that would cover it is perhaps tricky).

js/components/dashboard/dashboard.js
```diff
@@ -120,21 +121,29 @@ export default class Dashboard extends PureComponent {
...
     } = this.props
     const anyStudentExpanded = expandedStudents.includes(true)
     const anyQuestionExpanded = expandedQuestions.includes(true)
     const showTallQuestionHeader = anyStudentExpanded || anyQuestionExpanded
     const activitiesList = activities.toList().filter(activity => activity.get('visible'))
     return (
       <div className={css.dashboard}>
+        <QuestionDetails
+          selectedQuestion={selectedQuestion}
+          onClose={(e) => selectQuestion(Map())}
```
pjanik: This is my typical React comment - I try to avoid arrow expressions in the render method. A new function is created each time render() is called, so QuestionDetails will be re-rendered each time too, even if its props are not changed. Most likely not causing any issues here, but personally I'd avoid this in general - to keep consistent style and one day it can cause real issues.

js/components/dashboard/question-details.js
```diff
@@ -0,0 +1,53 @@
...
+    const answers = selectedQuestion && selectedQuestion.get('answers')
+      ? selectedQuestion.get('answers').sortBy(a =>
+        (a.getIn(['student', 'lastName']) + a.getIn(['student', 'firstName'])).toLowerCase())
+      : []
+    const prompt = selectedQuestion && selectedQuestion.get('prompt')
+    return (
+      <Modal
+        show={selectedQuestion.size > 0}
+        onHide={onClose}
+        contentLabel='Example Modal'
```
pjanik: Not sure if we want "Example modal" label. :wink: And actually I've seen some warning in dev console re contentLabel prop.

js/reducers/dashboard-reducer.js
```diff
@@ -2,14 +2,16 @@ import { Map } from 'immutable'
...
+  SELECT_QUESTION
 } from '../actions/dashboard'
 
 const INITIAL_DASHBOARD_STATE = Map({
   sortBy: SORT_BY_NAME,
   expandedActivities: Map(),
   expandedStudents: Map(),
-  expandedQuestions: Map()
+  expandedQuestions: Map(),
+  selectedQuestion: Map()
```
pjanik: Is this Map() a map of selected questions or just a question state?
If it a single question, I'd perhaps expect `null` here or ID instead of a copy of data. Copying can have some side effects if the data is updated and they will be out of sync. And that can actually happen, as the dashboard is meant to be used in "live" mode, so the new answers can show up.

knowuh:
@pjanik thanks for the review.

I believe that  a746459 and bbdc088 Address your valid concerns.
Most of the changes were pretty simple, switching from a copy of the selected question object to a live reference was a little tougher, but I could easily reuse and extend some selectors in the report. 

When you get a chance, please review the two commits mentioned above.  🙏

js/selectors/dashboard-selectors.js
```diff
@@ -1,10 +1,25 @@
...
+// When a user selects a to display question details by
+// Clicking on the question column expand icon.
+export const getSelectedQuestion = createSelector(
+  [getQuestionTrees, getSeletedQuestionId],
+  (questions, id) => {
+    if (questions && id) {
+      return questions
+        .get(id.toString(), fromJS({}))
+    }
+    return fromJS({})
```
pjanik: Shouldn't that be `null`? Somewhere in the reducer, the initial value is `null`. I guess it works, though. ;)

knowuh: af1d6e9 should fix it! :)

PR #84, opened 2018-10-11
Fix negative value bug in maxScore.

knowuh:
+ add test for NumericTextField

Part of  this PT Story:

'Make default max score 10 instead of 0.'
[#130389079]
https://www.pivotaltracker.com/story/show/130389079

### DIFF
--- a/test/components/report/numeric-text-field_spec.js
+++ b/test/components/report/numeric-text-field_spec.js
@@ -47,4 +47,10 @@ describe('<NumericTextField /> minValue prop', () => {
     wrapper.find('input').simulate('blur', {})
     expect(wrapper.state('value')).to.equal(-3)
   })
+  it('wont allow zero value when minimum is 1; with default 10', () => {
+    const wrapper = shallow(<NumericTextField default={10} min={1} />)
+    wrapper.find('input').simulate('change', { target: { value: '0' } })
+    wrapper.find('input').simulate('blur', {})
+    expect(wrapper.state('value')).to.equal(10)
+  })
 })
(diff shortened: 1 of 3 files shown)


test/components/report/numeric-text-field_spec.js
```diff
@@ -0,0 +1,50 @@
...
+    wrapper.find('input').simulate('blur', {})
+    expect(wrapper.state('value')).to.equal(10)
+  })
+  it('will not allow negative numbers; default to default value', () => {
+    const wrapper = shallow(<NumericTextField default={10} />)
+    wrapper.find('input').simulate('change', { target: { value: '-1' } })
+    wrapper.find('input').simulate('blur', {})
+    expect(wrapper.state('value')).to.equal(10)
+  })
+  it('will not allow numbers less than default (10) as we specify', () => {
```
DALove1025: I think the "it" description might be confusing. To me, this is testing that `it('will allow...'`

test/components/report/numeric-text-field_spec.js
```diff
@@ -0,0 +1,50 @@
...
+    wrapper.find('input').simulate('change', { target: { value: '0' } })
+    wrapper.find('input').simulate('blur', {})
+    expect(wrapper.state('value')).to.equal(0)
+  })
+  it('will allow negative values; with default 10, and min as -10', () => {
+    const wrapper = shallow(<NumericTextField default={10} min={-10} />)
+    wrapper.find('input').simulate('change', { target: { value: '-3' } })
+    wrapper.find('input').simulate('blur', {})
+    expect(wrapper.state('value')).to.equal(-3)
+  })
```
DALove1025: I'm wondering if there's one more test of value -- `it('will allow an in-range decimal value; rounds to nearest integer')` -- test.    Set the value to 2.6 and have `expect(wrapper.state('value')).to.equal(2)`

PR #86, opened 2018-10-29
Handle case where error occurs while saving teacher feedback

scytacki:
Also handle network errors that occur during fetches
[#128503731]


pjanik:
@scytacki, I've pushed a small fix that should handle this case. You can double check if it works, as I guess that was the case that caused all the initial troubles. Network error should be fine once the network is restored (only part of the data would be missing).

PR #87, opened 2018-10-31
Test teacher feedback

scytacki:
This test feels messy, but I learned a lot about cypress doing it.
And in the end it does verify that this commit is needed:
https://github.com/concord-consortium/portal-report/commit/1c7d2ae11d8c3cd7f8c03ab3e4faa6075ac5d6b9

### DIFF
--- a/cypress/integration/feedback.spec.js
+++ b/cypress/integration/feedback.spec.js
@@ -1,0 +1,51 @@
+/* global describe, it, beforeEach, cy */
+/* eslint-disable-next-line */
+import React from 'react'
+import fakeData from '../../js/data/report.json'
+import sampleRubric from '../../public/sample-rubric'
+
+describe('Provide Feedback', function () {
+  beforeEach(() => {
+    // cypress currently doesn't handle window.fetch so:
+    // diable browser's fetch this way the fetch polyfill is used which uses XHR
+    cy.on('window:before:load', win => {
+      win.fetch = null
+    })
+
+    cy.server()
+
+    // Return the fake report data
+    cy.route({
+      method: 'GET',
+      url: '/',
+      response: fakeData
+    })
+
+    // Fake report data references sample-rubric.json file, so it has to be stubbed too
+    cy.route({
+      method: 'GET',
+      url: '/sample-rubric.json',
+      response: sampleRubric
+    })
+
+    // On first load portal-report does a PUT request to save some report settings
+    cy.route({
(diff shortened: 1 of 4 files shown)


cypress/integration/feedback.spec.js
```diff
@@ -0,0 +1,80 @@
+/* global describe, it, beforeEach, cy, expect */
+/* eslint-disable-next-line */
+import React from 'react'
+import fakeData from '../../js/data/report.json'
+import sampleRubric from '../../public/sample-rubric'
+
+describe('Provide Feedback', function () {
+  beforeEach(() => {
+    // cypress currently doesn't handle window.fetch so:
+    // diable browser's fetch this way the fetch polyfill is used which uses XHR
```
kswenson: "disable"

cypress/integration/feedback.spec.js
```diff
@@ -0,0 +1,80 @@
...
+
+    // On first load portal-report does a PUT request to save some report settings
+    cy.route({
+      method: 'PUT',
+      url: '/',
+      response: {}
+    }).as('putReportSettings')
+
+    let fakeServer = 'http://portal.test'
+    cy.visit(`http://localhost:8080/?reportUrl=${encodeURIComponent(fakeServer)}`)
```
pjanik: I think we shouldn't provide a hostname in the `visit` calls, so tests don't depend on dev setup and so it's easier to test against staging or production too. We can set `baseUrl` option in cypress.json file and then just call:
```cy.visit(`/?reportUrl=${encodeURIComponent(fakeServer)}`)```
Then, I can use CYPRESS_BASE_URL env variable to point it to my local server (I'm using 8081 port), staging or production. Or we can provide some more fancy configuration in the future (like here https://github.com/concord-consortium/lara-cypress-tests/pull/2).

File: cypress/integration/feedback.spec.js (hunk 1 above)
Comment (scytacki) [reply cypress/integration/feedback.spec.js]: @kswenson have you seen the new suggestion option in GitHub, this is a perfect example for it. When making a comment it is the icon that looks like two sheets of paper and a +/- sign.

```suggestion
    // disable browser's fetch this way the fetch polyfill is used which uses XHR
```

ekosmin: Looks like this typo snuck back in the last commit :P 

```suggestion
    // disable browser's fetch this way the fetch polyfill is used which uses XHR
```

PR #88, opened 2019-02-15
Simple fix to ensure expanded content fits within the window viewport

dstrawberrygirl:
There's a lot going on in the code to handle scrolling in different places so I've taken a light approach here to ensure that the content should all be visible and scrollbars should always appear in line with request from Trudi.

The addition of a timeout is undesirable - but without it, the bottom of the vertical scroll panel will be flush with the bottom of the page until the page is resized, then it jumps up a small amount when resized. I think @pjanik you worked on some of the scrolling and calc logic in the past, would be great to know if you can see a more elegant solution!

### DIFF
--- a/js/components/dashboard/dashboard.js
+++ b/js/components/dashboard/dashboard.js
@@ -38,6 +38,7 @@ export default class Dashboard extends PureComponent {
   }
   componentDidUpdate(prevProps) {
     if (!_timeout && prevProps.expandedStudents !== this.props.expandedStudents) {
+      // timer used for fixing vertical container resizing after expanded data is displayed
       _timeout = window.setTimeout(this.onResize, 2000);
     }
   }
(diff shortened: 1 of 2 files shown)


js/components/dashboard/dashboard.js
```diff
@@ -34,6 +36,12 @@ export default class Dashboard extends PureComponent {
     this.activityHeaders.addEventListener('scroll', this.onHeadersScroll)
     this.onResize()
   }
+  componentDidUpdate(prevProps) {
```
knowuh: I am not convinced we need to do anything special here in `componentDidUpdate`.  

I didn't see any glitchy or weird behavior without it. 

I leave it to @dstrawberrygirl and @pjanik to figure out the best path if anything needs to be done.

pjanik:
I ran "GeniGUIDE Teacher Workshop 2018: Geniventure Pre-Quiz (2018-2019)" mentioned by Trudi using production and this branch, and frankly speaking, I didn't notice any difference on my mac (comparing side by side). Scroll bars are still hidden. Also, because of that, I can't comment on this forced resizeHandler.

What browser and OS are you using @dstrawberrygirl?
But I guess it might not help Trudi if I don't see any changes.

scytacki:
This was fixed a different way.

PR #90, opened 2019-03-03
Google analytics

emcelroy:
This adds Google Analytics for tracking user interactions with the report and dashboard interfaces.


emcelroy:
@dougmartin: Thanks. Does the latest commit add what you were suggesting?

PR #97, opened 2019-04-29
Editing form

knowuh:
This is the first pass at trying to offer a better experience for authoring Rubric.

It's only marginally better than markdown inside of JSON.  :) I didn't have time to clean this up as much as I would have liked, &etc.

Under the hood it uses:

* [Formik](https://jaredpalmer.com/formik/)
* [React Tabs](https://github.com/reactjs/react-tabs)
* [React MDE](https://github.com/andrerpena/react-mde)

I tried using [React Schema Form](https://github.com/mozilla-services/react-jsonschema-form), but I failed to make suitable progress quickly.  It might be worth trying again, in which case there is also an online  tool which can infer JSON schema from examples. [Schema Tool](https://jsonschema.net/) It does a reasonable job.

## Important Notes:

* Authors must click on the green `Update` button at the top left of the form in order for the other tabs (JSON | TEACHER | STUDENT) to update their previews.
* I might have also broken the ability to edit the form directly in JSON.  This shouldn't be hard to fix, I just ran out of time.


knowuh:
It looks like tests are failing, probably due to react update.

knowuh:
Now it looks as though Cypress is having trouble running on Travis.   The error is 


`/home/travis/.cache/Cypress/3.2.0/Cypress/Cypress: error while loading shared libraries: libgconf-2.so.4: cannot open shared object file: No such file or directory`

knowuh:
I am closing this one -- its a bit dirty. Opening two separate PRs   instead.

PR #101, opened 2019-05-08
Add rubric migrations and test for migrations.

knowuh:
While fixing [#165857915]  "An author can not uncheck non-applicable rating"

https://www.pivotaltracker.com/n/projects/736901/stories/165857915

This migration generates an verbose representation of the `nonApplicableRatings` key of the `criteria` objects. The value is an array of strings which should match the ratings keys. In the past this was a sparse array of key names. Now it is redundantly including a string for every rating, using the rating id as a value if the rating should be skipped.

This is a confusing format, and only temporary while Kiley gets her work done.

I am working on the next migration, but it will not be backwards compatible, and will use the rating ids as boolean properties of the `nonApplicableRatings` values.  It will be in its own branch and PR because I don't want the authors using a schema that isn't available on production reports.

### DIFF
--- a/public/sample-rubric.json
+++ b/public/sample-rubric.json
@@ -1,7 +1,7 @@
 {
     "id": "RBK1",
-    "formatVersion": "1.0.0",
-    "version": "12",
+    "version": "1.0.0",
+    "versionNumber": "12",
     "updatedMsUTC": 1519424087822,
     "originUrl": "http://concord.org/rubrics/RBK1.json",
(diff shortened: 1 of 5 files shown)


public/sample-rubric.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "RBK1",
-    "formatVersion": "1.0.0",
-    "version": "12",
+    "version": "1.0.0",
+    "versionNumber": "12",
```
knowuh: `versionNumber` is meant to track the number of times this particular piece of rubric has been modified.  We aren't really using it, but it was part of the original discussion about rubric work. 

I think it could probably be removed.

PR #103, opened 2019-05-16
Major maintenance: TypeScript support, Mocha => Jest, StandardJS => TSLint, linting, dependencies update

pjanik:
This is a pretty major update of this repo to make it more consistent with our other recent TS projects.

1. I've added unobtrusive TypeScript support. You can start writing TS but you don't have to. JS and TS can be mixed. I've converted a few files to prove that it works (e.g. config.js => config.ts and some other tiny module).

2. I replaced StandardJS linter with TSLint and our typical config. TSLint will lint both TS and JS files. I've converted all the source to a new style and fixed all the lint errors.

3. I've converted unit tests from Mocha to Jest. First, because the old setup was using Babel to transpile and I wanted to get rid of Babel. Second, we use Jest in most projects.

4. I've updated pretty much all the dependencies to the latest versions (including webpack and its config).


dougmartin:
Nice work - this should make portal pages much more maintainable.

PR #104, opened 2019-05-23
Support new activity structure JSON, render blank page, update tests

pjanik:
It might be a bit difficult to go file by file, so summary of changes:

1. I've worked with fake data so far. There's a new file in `js/data/sequence-structure.json`. It's basically JSON exported from LARA of my test sequence, with IDs added and a few changes (children + type instead of named children).
2. Renamed Investigation to Sequence, "fetchData" to "fetchReportStructure", tried to cleanup a few things.
3. Simplified transform-json-response helpers. Normalization is simpler now and it pretty much flattens report structure only. It doesn't touch students, answers, visibility filters, and feedback (they will come in a very different API responses).
4. Updated code to work with the new format, names, structures.
5. Made sure that report renders correctly activity structure. So, sequence, activity, and page names, question names and prompts. For multiple-choice question, it also renders correct summary.
6. Question selection and visibility filter work client-side (I didn't focus on API, as this part will be reworked totally).
7. Dashboard renders too.
8. I've updated most of the broken tests, so they actually still work and have value. I only skipped a few of them that specifically required answers to be available to calculate progress (there're TODO comments and `x` prefix in front of Jest's `describe` function => `xdescribe`).
9. I disabled Cypress tests on Travis, as obviously, they can't work now. But it will be a nice test at the end of this refactoring process.

Demo:
http://portal-report.concord.org/branch/165518493-blank-page/index.html


js/core/transform-json-response.js
```diff
@@ -1,99 +1,72 @@
...
+  // Add some question properties, e.g. question numbers, selection, visibility.
+  resourceJson.children.forEach(activity => {
+    activity.children.forEach(section => {
+      // Questions are numbered within the whole activity.
+      let currentQuestionNumber = 1;
+      section.children.forEach(page => {
+        page.children.forEach(question => {
+          question.questionNumber = currentQuestionNumber++;
+          // Id is not enough, as questions of different type can have the same id.
+          question.key = `${question.type}-${question.id}`;
```
knowuh: Do you think we should create this id on export from LARA?

js/core/transform-json-response.js
```diff
@@ -1,99 +1,72 @@
...
+          // Nothing is selected by default.
+          question.selected = false;
+          if (question.type === "multiple_choice") {
+            // Multiple choice question is scored if at least one choice is marked as correct.
+            question.scored = question.choices.some(c => c.correct);
```
knowuh: Same question here:  Should we export the 'scored' question property as part of the new schema?

scytacki: Having a scored option would enable us to support an multi-select question that had no correct answers. That seems like a good thing. LARA doesn't support this now, but it seems good to make it explicit.

js/core/transform-json-response.js
pjanik: I guess it's fine that we expect authoring system / LARA to publish things that have unique `type + id` combination and deal with that here (as it could also affect other containers with clashing IDs and different types).

But I'm wondering how will we match answers and questions. Have you thought about that in your LARA run export code? 
I guess it should be also something like this `key` here. Then, it could make sense for LARA to generate these keys both in questions and answers, so we can be sure that they match and reporting system doesn't have to care about that.

js/core/transform-json-response.js
pjanik: Hm, I'm 50-50 here. I feel like `scored` can be always derived from `correct` properties. So LARA in fact already provides this information. This logic here is just simplification for UI components, so they don't have to do this check all the time. 
Is there a case when the question is scored, but none of the choices is considered to be correct?

@scytacki, I'm not sure if I understand your use case and how `scored` would relate to multi-select questions. Note that here we talk about question metadata - whether a given choice is correct. It's not an answer and whether a choice is selected (if that's you meant).

pjanik:
I've replied to some questions. I guess the question `key` might be important.

I've also renamed some types (Sequence, Activity, Section, Page) to lowercase, so they match question types.

js/core/transform-json-response.js
knowuh: I   think you are right: Lara should create this key in questions and answers (for matching  purposes)

js/core/transform-json-response.js
scytacki: For a normal multiple choice question the student has to choose one and only one answer. Once they select a answer there is no way to unselect all of the answers.  So if it is scored, it will always have a correct answer.

For a multiple select multiple choice. The student could choose 0 to N answers. And even if they choose 1 answer they can still unselect it.  So in theory choosing 0 could be the correct way to respond to the question.  For example: "Select all of the items that are true." and then none of the choices are actually true. So the correct response is to select nothing. But now that I think about this, we wouldn't know the difference between the student skipping the question and answering it. So in practice this is probably never used. There would have to be a choice like "none of the answers are true". 

So based on this I think you are right there isn't a valid use case where an answer is scored and there are no correct answers.

pjanik: Ah, I understand it now. Right, I think being explicit and providing "none of the answers are true" option makes more sense.

pjanik:
Ok, so to resolve comments above: I've assumed that "key" and "question_number" is coming from LARA (and updated LARA to send these values too). So, I'm going to merge this PR.

PR #106, opened 2019-05-30
Sign in to Firebase, fetch more data from Portal - class, students, JWT

pjanik:
[#165939200]
[#165518667]

Bunch of refactoring related to new data format and sources. Instead of fetching just one report JSON, we now fetch:
- offering data
- class info
- firebase JWT
Then we sign in to Firestore, and fetch/observe activity structure.

All that has been implemented in this PR. Dummy data is still supported, although it's getting a bit more complex because of the number of requests. I've tried to test various cases when something fails and make sure we still display a reasonable error message (e.g. firebase auth issue).

Students are listed in Dashboard, but still not in the regular report view as that's bound to answers there - this work is in progress.

I've also removed the refresh button and some logic related to it (including the timestamp). Now, everything will be "live", so there's no need for that.


js/core/transform-json-response.ts
```diff
@@ -0,0 +1,161 @@
...
+  children: IResource[] | IQuestion[];
+}
+
+export interface IQuestion extends IResource {
+  prompt: string;
+  selected: boolean;
+  scored?: boolean; // multiple choice only
+  choices?: IChoice[]; // multiple choice only
+}
+
```
knowuh: It looks really nice to see this typed.  👏

PR #110, opened 2019-06-12
Use real class_hash, fix compare view, dashboard, fix and update specs

pjanik:
1. Real class_hash is being used now.
2. Compare answers view and other modal dialogs were not functional in the new version of the teacher report. It's fixed now.
3. The dashboard view is also fixed (that was the biggest change in this PR).
4. Dashboard tests are updated, some new are also added.
5. Other random bug fixes.


js/components/dashboard/activity-answers.js
```diff
@@ -1,10 +1,13 @@
 import React, { PureComponent } from "react";
 import ProgressBar from "./progress-bar";
-import Answer from "./answer";
+import Answer from "../../containers/dashboard/answer";
 
 import css from "../../../css/dashboard/activity-answers.less";
 
 export default class ActivityAnswers extends PureComponent {
+
+  studentAnswer;
```
knowuh: Is this an unused property? 

Possibly a typo.

test/selectors/dashboard-selectors_spec.js
```diff
@@ -87,15 +84,15 @@ describe("dashboard selectors", () => {
...
     // TODO fix these tests when we start supporting answers using new data format
-    xdescribe("when sorting by most progress", () => {
+    describe("when sorting by most progress", () => {
       it("should return sorted list of students", () => {
         expect(getSortedStudents(state({sortBy: SORT_BY_MOST_PROGRESS})).toJS()).toEqual(// Students sorted by most progress (ties broken alphabetically)
         [ s2, s1, s3 ]);
       });
     });
 
     // TODO fix these tests when we start supporting answers using new data format
```
knowuh: Can we get rid of these TODO items now?

js/components/dashboard/activity-answers.js
pjanik: Thanks, fixed now.

test/selectors/dashboard-selectors_spec.js
pjanik: Right, they are removed now.

PR #115, opened 2019-06-25
Hot-fix: CSS (boostrap modal) regression in CompareView

knowuh:
bootstrap and react-bootstrap were updated along with React 16.
Animation interfaces must have broken, because faded components never
un-fade.

This fix is a little bit of a hack because it doesn't restore animation, it just makes
sets faded bootstrap components to opaque instead of transparent.

https://www.pivotaltracker.com/story/show/166907544
[#166907544] fix-compare-view

Compare view no longer works.

This is a regression from v1.9.0 to v1.10.0, probably a result of updating React to 16.

### DIFF
--- a/css/report/bootstrap-components.less
+++ b/css/report/bootstrap-components.less
@@ -10,7 +10,6 @@
 .modal-content {
   color: @color;
   font: @font;
-
   a {
     color: #479492;
     text-decoration: none;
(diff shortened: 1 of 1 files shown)


pjanik:
Firestore-based version already fixed that and transitions work there. Well, actually by restoring older bootstrap version (not sure if there's big value in using a newer one, since that's the only thing from Bootstrap that we use AFAIR). This fix looks good to me, but I don't think it should be merged into master. I guess we should just release it as v1.10.1.

knowuh:
Closing this PR because it should NOT be merged into master.

PR #116, opened 2019-06-28
[#166240880] FireStore: Save assignment level report settings

knowuh:
We save the teachers settings for the offering at this FireStore path:
/sources/${sourceId}/user_settings/${platformUserId}/offering/${contextId}

We dispatch all user settings updates to FireStore via the middleware API.
The RECEIVE_USER_SETTINGS action is the one which changes the active store.

Even when the network is disabled (wifi disconnected) the FireStore API dispatch method
works, and the RECEIVE_USER_SETTINGS is reliably called with valid parameters.

The FireStore rules are setup to allow unfettered access to:
`/sources/fake.portal/user_settings/1/offering/class123`

This allows us to test the report settings with fake portal data, without a valid JWT.

https://www.pivotaltracker.com/story/show/166240880

### DIFF
--- a/js/api.ts
+++ b/js/api.ts
@@ -11,9 +11,10 @@ const FIREBASE_APP = "report-service-dev";
 const FAKE_FIRESTORE_JWT = "fake firestore JWT";
 
 export interface ILTIPartial {
-  platformId: string;
+  platformId: string;      // portal
   platformUserId: string;
-  contextId: string; // class hash
+  contextId: string;       // class hash
+  resourceLinkId?: string;  // offering ID
 }
 
 export interface IPortalRawData extends ILTIPartial{
@@ -167,23 +168,20 @@ export function fetchPortalDataAndAuthFirestore(): Promise<IPortalRawData> {
 }
 
 export function reportSettingsFireStorePath(LTIData: ILTIPartial) {
-  const {platformId, platformUserId, contextId} = LTIData;
+  const {platformId, platformUserId, resourceLinkId} = LTIData;
   const sourceId = platformId.replace(/https?:\/\//, "");
   // NP: 2019-06-28 In the case of fake portal data we will return
   // `/sources/fake.portal/user_settings/1/offering/class123` which has
   // special FireStore Rules to allow universal read and write to that document.
   // Allows us to test limited report settings with fake portal data, sans JWT.
-  return `/sources/${sourceId}/user_settings/${platformUserId}/offering/${contextId}`;
+  return `/sources/${sourceId}/user_settings/${platformUserId}/resource_link/${resourceLinkId}`;
 }
 
 export function updateReportSettings(update: any, state: ILTIPartial) {
   const path = reportSettingsFireStorePath(state);
-  return new Promise( (resolve, reject) => {
-    firebase.firestore()
+  return firebase.firestore()
       .doc(path)
       .set(update, {merge: true});
-    resolve();
-  });
 }
 
 // The api-middleware calls this function when we need to load rubric in from a rubricUrl.
--- a/js/api.ts
+++ b/js/api.ts
@@ -1,3 +1,4 @@
+import * as firebase from "firebase";
 import fetch from "isomorphic-fetch";
 import * as jwt from "jsonwebtoken";
 import fakeOfferingData from "./data/offering-data.json";
@@ -9,7 +10,13 @@ import * as db from "./db";
 const FIREBASE_APP = "report-service-dev";
 const FAKE_FIRESTORE_JWT = "fake firestore JWT";
 
-export interface IPortalRawData {
+export interface ILTIPartial {
+  platformId: string;
+  platformUserId: string;
+  contextId: string; // class hash
+}
+
+export interface IPortalRawData extends ILTIPartial{
   offering: {
     id: number,
     activity_url: string;
@@ -20,9 +27,6 @@ export interface IPortalRawData {
     students: IStudentRawData[];
   };
   userType: "teacher" | "learner";
-  platformId: string;
-  platformUserId: string;
-  contextId: string; // class hash
 }
 
 export interface IStudentRawData {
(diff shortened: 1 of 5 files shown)


knowuh:
@pjanik could you please take a look?

js/reducers/index.js
```diff
@@ -174,14 +172,19 @@ function report(state = INITIAL_REPORT_STATE, action) {
...
+    // to firebase. The results of the calls are handled by RECEIVE_USER_SETTINGS
+    case SET_ANONYMOUS:
     case HIDE_UNSELECTED_QUESTIONS:
-      return hideUnselectedQuestions(state);
     case SHOW_UNSELECTED_QUESTIONS:
-      return showUnselectedQuestions(state);
-    case SET_ANONYMOUS:
-      return setAnonymous(state, action.value);
+    case SET_QUESTION_SELECTED:
+      return state; // These ⬆ API-calling actions just return the current state.
```
pjanik: I think we don't need to list these actions here, as they don't do anything. `default` handler simply returns state.
But I don't understand how that works now - does it mean that once a user selects a question, it will be locally selected only when we get a response from Firestore (using RECEIVE_USER_SETTINGS handler)? Then it'd depend on network conditions, so I'm not sure about that.

js/reducers/index.js
```diff
@@ -63,34 +65,27 @@ function data(state = INITIAL_DATA, action) {
...
-  return state.set("anonymous", anonymous)
-    .set("students", newStudents);
+  const newStudents = state.get("students")
+    .map(s => s.set("name", anonymous ? `Student ${idx++}` : s.get("realName")));
+  return state.set("anonymous", anonymous).set("students", newStudents);
 }
 
-// This action has to be explicit and requires additional property. Otherwise, a question will disappear immediately
-// when teacher has selection filter active and unselects given question. Instead, the question should stay visible
-// until teacher clicks "Show selected" again.
```
pjanik: Note that you lost this behavior described in this comment. Selecting a question immediately triggers filtering, which feels a bit strange, as other questions disappear and you cannot select them anymore. You can compare:
http://portal-report.concord.org/branch/166240880-assignment-level-report-settings/index.html
and
https://portal-report.concord.org/version/v1.7.1/index.html

js/api.ts
```diff
@@ -162,9 +166,24 @@ export function fetchPortalDataAndAuthFirestore(): Promise<IPortalRawData> {
...
-  // TODO using Firebase
-  return new Promise(resolve => resolve());
+export function reportSettingsFireStorePath(LTIData: ILTIPartial) {
+  const {platformId, platformUserId, contextId} = LTIData;
+  const sourceId = platformId.replace(/https?:\/\//, "");
+  // NP: 2019-06-28 In the case of fake portal data we will return
+  // `/sources/fake.portal/user_settings/1/offering/class123` which has
+  // special FireStore Rules to allow universal read and write to that document.
+  // Allows us to test limited report settings with fake portal data, sans JWT.
+  return `/sources/${sourceId}/user_settings/${platformUserId}/offering/${contextId}`;
```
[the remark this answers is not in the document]
knowuh: Good point about the documentation.  
Perhaps I can initialize the store using local JSON.

js/reducers/index.js
knowuh: Ah, right!  I didn't immediately understand this.  

Thanks.

js/reducers/index.js
knowuh: Because the `RECEIVE_USER_SETTINGS` call is a side-effect of the API middleware invocation, I wanted to put documentation of that in the actions handler. 

This approach works when detached from the network.  `RECEIVE_USER_SETTINGS` is called with the data we expect. as if we were able to connect to FireStore:

> Local writes in your app will invoke snapshot listeners immediately. This is because of an important feature called "latency compensation." When you perform a write, your listeners will be notified with the new data before the data is sent to the backend.  [link](https://firebase.google.com/docs/firestore/query-data/listen)

knowuh:
@pjanik thank you for the feedback.

166aaa4 addresses the big concerns you brought up, and things I missed.

I hope the discussions above will clear up other questions.  I am off until Monday, so review at your leisure.

js/reducers/index.js
knowuh: I restored this functionality starting on line 89 below.

js/reducers/index.js
```diff
@@ -174,14 +197,23 @@ function report(state = INITIAL_REPORT_STATE, action) {
       return state.set("answers", Immutable.fromJS(preprocessAnswersJSON(action.response)));
     case SET_NOW_SHOWING:
       return state.set("nowShowing", action.value);
-    case SET_QUESTION_SELECTED:
-      return state.setIn(["questions", action.key, "selected"], action.value);
+
+    // The following actions only trigger API middleware hooks that call out
```
pjanik: I guess this comment is outdated now since these functions call something else than middleware?

js/api.ts
```diff
@@ -162,9 +166,24 @@ export function fetchPortalDataAndAuthFirestore(): Promise<IPortalRawData> {
...
+}
+
+export function updateReportSettings(update: any, state: ILTIPartial) {
+  const path = reportSettingsFireStorePath(state);
+  return new Promise( (resolve, reject) => {
+    firebase.firestore()
+      .doc(path)
+      .set(update, {merge: true});
+    resolve();
```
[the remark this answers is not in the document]
pjanik: Seems to be fixed, thanks.

js/reducers/index.js
pjanik: Okay, that makes sense. And it's nice, since it can simplify initial settings initialization and update process, as everything is handled just by one `RECEIVE_USER_SETTINGS` handler.
Maybe I'd add some comment somewhere that this Firestore update resolves immediately, so it's clear.

js/reducers/index.js
pjanik: 👍

pjanik:
Thanks, it looks good now. I think I only pointed one outdated comment (if I'm right) and maybe I'd mention somewhere that Firestore action resolves immediately (as that's not typical for network-based actions), so we can use this approach that you've started.

knowuh:
@pjanik -- I forgot to ask you to please look at the comments I added in 
4d8b05f

pjanik:
Thanks, it looks good. 👍

PR #117, opened 2019-07-19
Do not modify the `platform_id` here in the report view code.

knowuh:
1. Technically it's OK to have trailing slashes in field values.
2. The platform is the authority on `platform_id` and it should be trusted here.

Full discussion can be found in the
[PT story](https://www.pivotaltracker.com/story/show/166241318)

[#166241318]


knowuh:
@scytacki  if you want to try your work with this branch `166241318-undo-slash-removing-platformid` it might be edcational.

PR #118, opened 2019-07-26
166241028 add firestore feedback

knowuh:
The feedback is now saved in Firestore using the same middleware API pattern that we used for the storing the report settings.

Also in this PR minor cleanup of string constants.

More work is needed to fully support this feedback. See [Feedback in New Reports #166241028](https://www.pivotaltracker.com/story/show/166241028) for details.


pjanik:
@scytacki, I've applied changes that we discussed on Slack to Noah's PR. Now question feedbacks are stored at: `/sources/<lara_hostname>/question_feedbacks/<answer_id>`.

Note that security rule looks like that:
```
match /sources/{source}/question_feedbacks/{document=**} {
  allow create: if request.auth.token.user_type == 'teacher';
  allow read, update: if request.auth.token.user_type == 'teacher' && request.auth.token.class_hash == resource.data.context_id;
  allow read: if request.auth.token.user_type == 'learner' && string(request.auth.token.platform_user_id) == string(resource.data.platform_user_id);
}
```

so it's a bit too open for `create`, but it seems that's the cost of the flat structure (I mentioned that at the beginning of the discussion).

Not sure if you want to review these changes, or should I just merge them now?

PR #121, opened 2019-08-09
Save and load Activity feedback to FireStore

knowuh:
This PR only gets the activity level feedback saving and loading again. [PT story](https://www.pivotaltracker.com/story/show/166241028)

The appropriate FireStore rules were added in this [report-service PR](https://github.com/concord-consortium/report-service/pull/9)

The biggest change is that feedback is no longer an array, because we only save the latest feedback.

Some remaining TODO items:

* Automatic activity level scores.  I took a look a this tonight, but didn't make much progress. The `makeGetQuestionAutoScores` inside of `activity-feedback-selectors.js` could probably be greatly simplified now that we have a much flatter feedack model.

* Activity Level feedback invalidation. We haven't thought much about this, but I believe that if a student does some more work, then the question feedback will be invalid, and we should then be able to invalidate our activity level feedback.  In this case, I think we might need an explicit 'has been reviewed flag on the record, or maybe we can do some similar hashing strategy.

* I haven't checked if activity level feedback is showing up on student reports yet.


js/selectors/activity-feedback-selectors.js
```diff
@@ -57,18 +59,15 @@ const activityFeedbackFor = (activity, student, feedbacks) => {
...
-  const feedback = fb.get("feedbacks") && fb.get("feedbacks").first();
-  return feedback && feedback.get("hasBeenReviewed");
+  return fb && fb.get("hasBeenReviewed");
 };
 
-const hasLearner = (feedback) => !!feedback.get("learnerId");
-
-const getFeedbacksNotAnswered = (fbs) => fbs.filter(f => !hasLearner(f));
+const getFeedbacksNotAnswered = (fbs) => fbs;
+// TODO: ⬆  How do we know if a student hasn't started yet?
```
pjanik: Right, I had the same problem when I started looking at this. I guess we should check if there's an answer for a given activity. We have this data available. Note that the dashboard has some selectors that calculate progress for each student, so maybe we could reuse it.

PR #124, opened 2019-08-15
Fix feedback statistics, show which students haven't started activity

pjanik:
[#167871598]

This PR fixes statistics in the feedback panel and ensures that students with no response are listed there. Note that it isn't based on master, but on the rubric-related branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ _site
 .env
 .idea
 .DS_Store
+coverage
 
 /cypress/screenshots
 /cypress/fixtures/example.json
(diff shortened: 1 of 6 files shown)


scytacki:
Looking at the error again, I now see that it did actually report the semicolon:
https://travis-ci.org/concord-consortium/portal-report/builds/572478218#L619

I wonder if there is some way to get `npm run lint` to stop printing the extra lines when it fails:
https://travis-ci.org/concord-consortium/portal-report/builds/572478218#L621-L630
They are distracting.

Looks like from this issue: https://github.com/npm/npm/issues/6124 we can remove those lines by running `npm run lint --silent` It sounds like that will also take away the log that prints the tslint command being run though which isn't great.

PR #125, opened 2019-08-21
Add support of `activityIndex` URL param

pjanik:
[#168002506]

### DIFF
--- a/js/core/transform-json-response.ts
+++ b/js/core/transform-json-response.ts
@@ -1,6 +1,7 @@
 import { normalize, schema } from "normalizr";
 import humps from "humps";
 import { IPortalRawData } from "../api";
+import queryString from "query-string";
 
 export interface IResource {
   id: number;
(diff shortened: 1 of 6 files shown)


js/core/transform-json-response.ts
```diff
@@ -116,6 +117,15 @@ export function preprocessResourceJSON(resourceJson: IResource) {
       });
     });
   });
+
+  // If `activityId` is provided as a URL parameter, filter activities to include only this one.
```
knowuh: I think that in the comment above `activityId` should be replaced with `activityIndex`?

pjanik: Yeap, thanks!

pjanik: Weird, I pushed a commit that fixed this comment, it was in a branch, but somehow PR didn't get updated. It's either Github having some issues or maybe this option that requires PR to be approved before it can be merged is causing that?
Anyway, I re-added it directly to master https://github.com/concord-consortium/portal-report/commit/bdea61e3b27544fc3cfc3b174e998a9e2c09e437

PR #129, opened 2019-09-13
Use url parameters to chose FireStore DB source.

knowuh:
This PR changes how an App initializes its firebase connection.

The new method looks for a `firebase-app` in the url parameters.  If it doesn't find one, it uses `report-service-dev` by default.

Because DB initialization happens later as a result. The `db` was modified, so that it wouldn't immediately initialize FireStore when loading. 

PT Story "Report should be configurable to run on Production"

[#168486280]
https://www.pivotaltracker.com/story/show/168486280

### DIFF
--- a/js/actions/index.ts
+++ b/js/actions/index.ts
@@ -1,4 +1,4 @@
-import db from "../db";
+import { db } from "../db";
 import fakeSequenceStructure from "../data/sequence-structure.json";
 import fakeAnswers from "../data/answers.json";
 import {AnyAction, Dispatch} from "redux";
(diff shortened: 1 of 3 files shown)


knowuh:
@scytacki  -- @dougmartin is sick, and @pjanik is done for the day, so I tagged you for review.

scytacki:
Oh it looks like the type checking has failed:
https://travis-ci.org/concord-consortium/portal-report/builds/584714750#L574-L596

scytacki:
👍

PR #132, opened 2019-10-11
Fix printing of student reports and student link in feedback panel

pjanik:
1. Portal Report no longer renders all the student reports hidden by CSS. This didn't work well when we had a class with 85 students.
2. Printing was broken in browsers other than Chrome. It seems API has changed over the years.
3. Student links in the feedback panel were pointing to a wrong activity if the original report was launched with the `activityIndex` URL param.


js/components/report/report.js
```diff
@@ -47,52 +62,53 @@ export default class Report extends PureComponent {
...
+    }
+    const selectedStudentIds = report.get("selectedStudentIds");
     let studentsToRender;
-    if (selectedStudentId) {
-      const selectedStudent = report.get("students").get(selectedStudentId);
-      if (selectedStudent) {
-        studentsToRender = [ selectedStudent ];
-      } else {
+    if (selectedStudentIds) {
+      studentsToRender = selectedStudentIds.map(sId => report.get("students").get(sId)).filter(s => !!s);
```
knowuh: I am afraid to ask what is happening here. 

Something very pathological about javascript boolean evaluation.  For what value is `!!a == true` but  `a == false`?   

Hold on while I go watch https://www.destroyallsoftware.com/talks/wat a few more times.

pjanik: Haha. 

>  For what value is !!a == true but a == false?
Never (I hope). 

I use `!!` to cast a value to `boolean`, e.g. values like `undefined` or `null` to `false` (even though they're "falsy" anyway, but TypeScript can complain about types). In this particular case `!!` is absolutely optional, I just want to filter out `undefined` values. `filter(s => s)` would work too. I just thought that casting to boolean will make it a bit more clear that this line/statement is not a mistake. Probably `s => s !== undefined` could work better for that.

PR #134, opened 2019-11-02
fix correct column in dashboard

scytacki:
PT Story: https://www.pivotaltracker.com/story/show/169513602

**Update 2019-11-15**: I'll leave the notes below for the history, here is an updated description:

The correct column is fixed by using a new selector which provides the  answers by question and then by student.  The ActivityAnswers component is now connected to the state to use this selector. This same selector is useful for several parts of the code which previously were iterating over the answers to find answers for a question or answers for a question and student. So that code has been updated. In some cases it was easier to continue to pass simple lists of answers arounds so I didn't updated those components.  Overall I think this has improved the performance, but we don't have a test to really verify that.  I did quickly test it with a large class in our production data. 

The PR also includes updated tests to cover the score column, and the completion status of required questions/answers. To test the completion status the mock or test data in `js/data` was updated to include a required question and a submitted and not submitted answer. The required questions/answers logic was touched as part of this work. All of the code that was modified as part of this work is covered by tests, however in some cases the tests might not be actually verifying the results match expectations.

If you are looking at the historic notes or commits, you can see I original added unit test for the scoring. I removed this test because I think it would be hard to maintain. The same code is covered by cypress tests.

**Historic Notes**
I'm not really sure the tests are worth it. They at least illustrate what state is used
to compute the correct answer. But I think they will be hard to maintain. The number
of references between objects based on ids, makes it really easy to have a mismatch
and then the tests fails for a very hard to find reason.

Also the tests are based on looking at what the code needs just to the right result. I actually fixed the problem by manually looking at the test data and debugging in the browser console. Then I added the tests to exercise the new code. 

A more useful test would be one that runs against the data in the js/data and checks
the result, but I don't think we have an example of that. Or perhaps there is a way to simplify the setup of the state.

Additionally, this new code will slow the dashboard down. It is already slow for large classes and this isn't going to help. I don't know enough about Redux and React to know what will be cached and what will be run multiple times. It seems like the `getAnswerTrees` selector is memoized so at least it doesn't run for each ActivityAnswers instance. But I'm pretty sure the  `mapStateToProps` function will run at least once for each instance. And for a class of 80 students with a sequence containing 6 activities. That is at least 480 runs. During each of these 480 runs it is searching through all of the answers of the class. So with 80 students, 6 activities, and 30 questions in each activity that is 14,400 answers if all of the students answered everything.  So that means 6,912,000 checks to see if an answer is should be included in the activities props of the ActivityAnswers component.  Ontop of this within each of these checks it is looping through the 30 questions of the activity to figure out if the answer is part of the current activity. So in the internal check inside of `questionIds.includes(answer.get("questionId"))` is run around 207,360,000 times (sometimes the loop will exit early, so that is a high estimate).


js/components/dashboard/activity-answers.js
```diff
@@ -51,3 +53,30 @@ export default class ActivityAnswers extends PureComponent {
...
+
+  return {
+    // This computes an answers property here by finding all of the answers
+    // for the questions we were passed.
+    // this is only used to compute a score, so it might be better to just compute
+    // the score here instead of passing around answers
+    answers: answerTreeList.filter((answer) => {
+      // This is not very efficient because we are iterating over every questionId in the
+      // the activity for each answer in the activity
+      return questionIds.includes(answer.get("questionId")) &&
```
kswenson: Would an initial pass over the `answerTreeList` to build a map from `questionId` to answers make sense? Or perhaps a map from `userId:questionId` to answers, which would be a larger map but involve less looping.

scytacki: Yes some kind of map here is what we need. I think ideally `getAnswerTrees(state)` would return a map or a map of maps so there is no need for filtering and iterating, just a simple lookup.  But I'm not sure if that is good practice. All of the selectors in https://github.com/concord-consortium/portal-report/blob/master/js/selectors/report-tree.js seem to return arrays.  In the example scenario above, that ought to cut down the 207,360,000 questionId checks to loop over 14,400 which creates maps.

scytacki:
I did some more code spelunking.  The result of the getAnswerTrees often filtered or iterated over. Here is how it is currently used:

- `components/dashboard/activity-answers`: answers that have a certain platformUserId and match a list of questionIds
- `containers/dashboard/answer`: answer with a platformUserId and questionId
- `containers/dashboard/question-details`: answers with questionId
- `containers/report/question`: answers with a questionId 
- `selectors/compare-view-data`: lookup by answer id
- `selectors/dashboard-selectors`: answer with questionId, platformUserId, and then checking its submitted property

Also it seems like getAnswerTrees is actually returning a Map from answer id to answer. That isn't obvious from the code but I guess `filter` and `map` work on Maps?   The only file using this map is `selectors/compare-view-data`.  And it could probably work on the raw state['answers'] map (I'm not sure what the correct notation is for that).

The rest of the usage would benefit from a map like: ```
{
  [questionId1]: {
    [platformUserId1]: answer1,
    [platformUserId2]: answer2,    
  },
  [questionId2]: {
    ...
  },
  ...
}
```
This would use more memory, but be faster most of the places it is used.

**Comment by scytacki** [comment]
A problem with this seems to be that the answers are modified (which I guess means a new answer is created since this is ImmutableJS), in the current getAnswersTrees. So the raw list of answers in the state will not work for `compare-view-data`. Perhaps a solution is provide two maps. So the result is actually:
```
{
  byAnswerId: {
    [answerId1]: answer1,
    [answerId2]: answer2,
    ...
  },
  byQuestionIdThenStudentId: {
    [questionId1]: {
      [platformUserId1]: answer1,
      [platformUserId2]: answer2,    
    },
    [questionId2]: {
      ...
    },
    ...
  }
}
```

scytacki:
I was wrong about not having an example of a test that used the mock data in js/data. The cypress tests run this way. So I'm going to add one for this case, and maybe I'll remove the other complex test.

scytacki:
I think this is mostly ready to go. I would like to change the name of `getAnswerTreesNew`.

scytacki:
Thanks Noah.

PR #136, opened 2019-11-06
Report testing

scytacki:


scytacki:
This has master merged in. And travis in enabled now, so you can see the test running there.

PR #138, opened 2019-11-09
Improve automated tests

scytacki:
This PR:
- refactors some of the cypress tests to remove redundant code
- fixes an issue with the travis config that was preventing it from sending the code coverage to code climate.
- disables firestore networking when the demo report is loaded
- fixes some cypress tests that were previously skipped because they were failing
- updates some cypress tests now that the user settings are starting empty on each test run

The most invasive change is the disabling of the firestore networking. This should only affect the demo report: the report that loads when you aren't coming from the portal.  Now the demo report won't remember any changes you made, but within the current tab session it should work as normal.

### DIFF
--- a/cypress/utils.js
+++ b/cypress/utils.js
@@ -1,9 +1,91 @@
+function getChildrenData(parent, label) {
+    let children = parent.children;
+
+    if (children != null) {
+        return children;
+    } else {
+        cy.log("There were no " + label);
+    }
+}
+
+function getByCypressTag ( cypressTag ) {
+  return cy.get(`[data-cy=${cypressTag}]`);
+}
+
+function getAnswerByQuestionType (answerData) {
+    let answer;
+    let questionType;
+    questionType = answerData.type;
+
+    if (answerData.type != null) {
+        switch (questionType) {
+            case ("Embeddable::MultipleChoice"):
+                answer = answerData.answer[0].choice;
+                break;
+            case ("Embeddable::OpenResponse"):
+                answer = answerData.answer;
+                break;
+            case ("Embeddable::ImageQuestion"):
+                answer = answerData.answer.image_url;
+                break;
+            case ("Embeddable::Iframe"):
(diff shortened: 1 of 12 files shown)


cypress/utils.js
```diff
@@ -1,9 +1,91 @@
...
+}
+
 module.exports = {
 
   // CSS modules randomize class names, so it is safest to select with explicit data-* attributes
   // https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements
-  getByCypressTag: function( cypressTag ) {
-    return cy.get(`[data-cy=${cypressTag}]`);
-  },
+  getByCypressTag: getByCypressTag,
```
knowuh: I think a valid simplification here is `module.exports = {getByCypressTag, getAnswerByQuestionType, getPageQuestionData, ... etc}`  

[Object Literal Property Value Shorthand](https://ariya.io/2013/02/es6-and-object-literal-property-value-shorthand)

cypress/utils.js
```diff
@@ -1,9 +1,91 @@
...
 
+  getAnswerByQuestionType: getAnswerByQuestionType,
+  getPageQuestionData: getPageQuestionData,
+  getActivityData: getActivityData,
+  getSectionData:  getSectionData,
+  getPageData:  getPageData,
+  getActivityQuestionData: getActivityQuestionData
 };
```
knowuh: I think a valid simplification here is `module.exports = {getByCypressTag, getAnswerByQuestionType, getPageQuestionData, ... etc}`  

[Object Literal Property Value Shorthand](https://ariya.io/2013/02/es6-and-object-literal-property-value-shorthand)

scytacki:
Thanks @knowuh for reviewing. I think I addressed all of your comments.

I'm not sure if the tests run faster. I've been using the code coverage support which makes them run slower. And I've been adding tests so that is changing the total test time.  My sense is that they run a little faster.

PR #139, opened 2019-11-20
fix feedback autoscore

scytacki:
The automatic total scoring which adds up the manual scores for individual questions was not reporting the correct value to the teacher. 

The cause of this problem was that the code was assuming a change of the value property on the NumericInputComponent would change the display. This component was mixing the patterns of controlled and uncontrolled components.  To fix the problem I made the NumericInputComponent be a explicitly uncontrolled. The parent components that use NumericInputComponent now have to deal with this uncontrolled approach. In some cases this is good because it makes the score behave the same as the text feedback box which is also uncontrolled. 

Bugs found while working on this:

- the wrong maxScore is saved and sent to the server when the teacher switches from manual to automatic activity scoring. I didn't have time to fix this issue, but added a comment.  I don't think the issue is critical though since the maxScore saved on the server should not be used when the activity scoring is automatic.
- the teacher assigned score is not saved until 1 second after the teacher stops typing in it. This slows down the tests, and it also could result in lost data if the teacher sets the score and moves on too quickly.  A fix for that isn't part of this PR, but I hope to create a new PR with a fix by using a underscore's debounce utility.
- the manually entered maxScore is not remembered if a teacher quickly switches to automatic scoring and then back again. There is a comment in the tests about this, it was not fixed in this PR.

This PR includes automated tests which verify the behavior of the places where the NumericTextField component is used. It also checks that the score is shown to the student correctly, which verifies that the score is making it through the firestore apis.

PT: [#169782026]

### DIFF
--- a/js/components/report/activity-feedback-options.js
+++ b/js/components/report/activity-feedback-options.js
@@ -51,6 +51,18 @@ export default class ActivityFeedbackOptions extends PureComponent {
       this.setState({lastScoreType: newV});
     }
     if (isAutoScoring(newV)) {
+      // When switching from manual to automatic scores, this sets the maxScore to the
+      // default Max Score (10), and then it doesn't seem to
+      // update this maxScore in the settings when the computation changes
+      // I suspect this is because the computedMaxScore prop is based on the old score type
+      // and it defaults to 10 when the score type is manual.
+      // Perhaps this code handles switching other modes better.
+      //
+      // TODO: tests for rubric, auto, manual, and no score to see if this
+      // line is doing something worth while in some case
+      //
+      // TODO: see if there is any reason why the computedMaxScore
+      // needs to actually be saved as the maxScore
       newFlags.maxScore = this.props.computedMaxScore;
     }
     updateActivityFeedbackSettings(activityId, activityIndex, newFlags);
--- a/js/components/report/score-box.js
+++ b/js/components/report/score-box.js
@@ -5,19 +5,35 @@ export default class ScoreBox extends PureComponent {
     super(props);
   }
 
+  // The score property is a little confusing here
+  // when the ScoreBox is disabled changes to the score property will be shown
+  // when it is not disable, the score property just initializes the text field.
+  // If you need the score to update you'll need to recreate the ScoreBox or switch it
+  // to disabled and then enabled.
+  renderScore(props) {
+    const {score, onChange, disabled } = props;
+    if (disabled) {
+      return <input
+        className="score-input"
+        value={score}
+        disabled="true"
+      />;
+    } else {
+      return <NumericTextField
+        className="score-input"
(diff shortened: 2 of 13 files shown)


scytacki:
- [ ]  check that the manually entered maxScore returns if the box is checked again this seems like it is working on the deployed master but locally it is not working for me.
  - [ ] address the TODOs in `activity-feedback-options.js` about saving the maxScore
  - [ ] document how improving the feedback box so it doesn't require a 1 second wait is not included in this PR
  - [ ] add tests for the max score being shown to students, testing the student view is part of a PR that is now in master so it should be easy to test the max score being shown to students.

js/components/report/score-box.js
```diff
@@ -5,19 +5,35 @@ export default class ScoreBox extends PureComponent {
     super(props);
   }
 
+  // The score property is a little confusing here
+  // when the ScoreBox is disabled changes to the score property will be shown
+  // when it is not disable, the score property just initializes the text field.
+  // If you need the score to update you'll need to recreate the ScoreBox or switch it
```
knowuh: If we wanted to update the `NumericTextField`'s  displayed value when its `intialValue` prop changed we could try using [`getDerivedStateFromProps(props, state)`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops).

But I don't see a scenario where this approach is going to be a problem ATM.

js/components/report/activity-feedback-options.js
```diff
@@ -51,6 +51,18 @@ export default class ActivityFeedbackOptions extends PureComponent {
...
+      // update this maxScore in the settings when the computation changes
+      // I suspect this is because the computedMaxScore prop is based on the old score type
+      // and it defaults to 10 when the score type is manual.
+      // Perhaps this code handles switching other modes better.
+      //
+      // TODO: tests for rubric, auto, manual, and no score to see if this
+      // line is doing something worth while in some case
+      //
+      // TODO: see if there is any reason why the computedMaxScore
+      // needs to actually be saved as the maxScore
```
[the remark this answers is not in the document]
scytacki: Yeah I don't think any of them have to happen right away. The two in this block are the same as they were before as far as I can tell. I think the bullets in the PR description summarize the TODO comments.

js/components/report/score-box.js
scytacki: That is kind of how it was trying to work before, but that seems to be a React antipattern since it is mixing a controlled and uncontrolled component. More info about that is in this slack thread:
https://concord-consortium.slack.com/archives/C0303QUNU/p1574100016370300
and this React post:
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

scytacki:
Thanks for the review @knowuh!

PR #142, opened 2019-12-10
use appropriate keys for feedback documents vs answer documents

scytacki:
Earlier I had refactored this code to reduce redundant logic when creating the collection queries. But I missed the fact the the answer documents and the feedback documents use different keys for the platform info. So this PR introduces extra logic to use the right keys.

I think the redundant refactoring is still worth it since it makes this inconsistency more visible Hopefully we can unify these keys in the future and remove the extra key logic.

### DIFF
--- a/js/actions/index.ts
+++ b/js/actions/index.ts
@@ -201,16 +201,16 @@ function watchFirestoreFeedbackSettings(db: firebase.firestore.Firestore, rawPor
 // Ugh the feedback system uses diffent keys than the answer system
 function correctKey(keyName: string, receiveMsg: string) {
   const feedbackKeys = {
-    "platform_id": "platformId",
-    "platform_user_id": "platformStudentId",
-    "resource_link_id": "resourceLinkId",
-    "context_id": "contextId"
-  }
+    platform_id: "platformId",
+    platform_user_id: "platformStudentId",
+    resource_link_id: "resourceLinkId",
+    context_id: "contextId"
+  };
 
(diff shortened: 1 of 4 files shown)


js/actions/index.ts
```diff
@@ -198,17 +198,42 @@ function watchFirestoreFeedbackSettings(db: firebase.firestore.Firestore, rawPor
   });
 }
 
+interface IStringMap {
```
dougmartin: You would get better type checking on the callers if you typed this code like this:

```
interface FeedbackKeys {
    platform_id: string;
    platform_user_id: string;
    resource_link_id: string;
    context_id: string;
}

const feedbackKeys: FeedbackKeys = {
    platform_id: "platformId",
    platform_user_id: "platformStudentId",
    resource_link_id: "resourceLinkId",
    context_id: "contextId"
};

export function correctKey(keyName: keyof FeedbackKeys, receiveMsg: string) { 
 ```

then correctKey("foo", "bar") wouldn't compile.

scytacki: Interesting, I didn't know about `keyof`. I think in this case using an array of possible keyNames would make more sense and then have feedbackKeys restrict its keys with that. But since this is intended as an internal method and will hopefully all go away, it doesn't seem worth it.

An tangential thing to note about this project is that when I run `npm run lint` it doesn't check the types as strictly as when I run `npm test`. I guess jest must be using a different typescript configuration than the lint command is using.

PR #143, opened 2020-03-12
Added getFirebaseJWT iframe-api endpoint [#171749310]

dougmartin:
Also fixed a runtime error (white page of death) when an activty has no questions.

### DIFF
--- a/js/components/report/interactive-iframe.js
+++ b/js/components/report/interactive-iframe.js
@@ -20,13 +20,18 @@ export default class InteractiveIframe extends PureComponent {
     };
     this.iframePhone = new iframePhone.ParentEndpoint(this.refs.iframe, phoneAnswered);
 
-    this.iframePhone.addListener("getFirebaseJWT", (options) => {
-      fetchClassData()
-        .then(classData => fetchFirestoreJWT(classData.class_hash, options.firebase_app))
-        .then(json => this.iframePhone.post("firebaseJWT", json))
-        // tslint:disable-next-line no-console
-        .catch(console.error);
-    });
+    this.iframePhone.addListener("getFirebaseJWT", this.handleGetFirebaseJWT);
+  }
+
+  handleGetFirebaseJWT = (options) => {
+    return fetchClassData()
+      .then(classData => fetchFirestoreJWT(classData.class_hash, options.firebase_app))
+      .then(json => {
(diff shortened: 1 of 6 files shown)


dougmartin:
@scytacki I didn't add tests as there are no existing tests of either the existing api.ts#fetchPortalDataAndAuthFirestore method where I pulled the two functions I needed for the new getFirebaseJWT endpoint or the interactive-iframe.js#InteractiveIframe component in either spec or Cypress tests.

scytacki:
I still think we add some tests. Even if they are simple like the fetchOfferingData test.

scytacki:
@dougmartin: ping to keep this on your radar

js/components/report/interactive-iframe.js
```diff
@@ -18,6 +19,14 @@ export default class InteractiveIframe extends PureComponent {
       this.iframePhone.post("initInteractive", typeof state === "string" ? JSON.parse(state) : state);
     };
     this.iframePhone = new iframePhone.ParentEndpoint(this.refs.iframe, phoneAnswered);
+
+    this.iframePhone.addListener("getFirebaseJWT", (options) => {
+      fetchClassData()
+        .then(classData => fetchFirestoreJWT(classData.class_hash, options.firebase_app))
+        .then(json => this.iframePhone.post("firebaseJWT", json))
```
scytacki: These lines aren't covered by a test.  I guess this will be tricky since you'll need to mock the embedding. Or mock iframePhone.

scytacki: If the mock report data interactive makes this getFirebaseJWT request, then this would probably be more covered by the existing cypress test. For the call to succeed though, the cypress tests would need to mock the request that is triggered by fetchFirestoreJWT but that shouldn't be hard using cypress' network mocking.

scytacki: @dougmartin just a ping on this. Perhaps we should give up on it at this point, but if you've got a little more patience for this, it seems like doing this before the Log Manager stuff would be good.

dougmartin: Will do today.

dougmartin:
@scytacki I was able to add jest tests for the getFirebaseJWT handler by extracting it out into a method on the class.  I've also verified that it loads the iframe with the branch as seen here:

image

PR #146, opened 2020-03-18
Fix: Activity order scrambled in teacher dashboard

knowuh:
The `toList()` call in the dashboard view made the assumption that the index order would be preserved in the conversion from map → list.
We need to manually sort on index after the conversion to a list.

I was trying to find a better place to do this sort, but its easy and clear to do it here in the `Dashboard` rendering.

[#170757052]
https://www.pivotaltracker.com/n/projects/736901/stories/170757052


scytacki:
A newer PR fixed this a different way.

PR #147, opened 2020-03-19
Better 🐞 fix  for Activity scrambled in dashboard

knowuh:
A better fix for 🐞 Activity order scrambled in teacher dashboard
[#170757052]

The previous PR sorted the activities (from activityTree) by the position attributes.

That sorting is not required if we use the `getSequenceTree()` Selector which is careful to
map deep references to children by `id` from the shallow normalized representation of the children array.

`report-tree_spec.js` was modified to use keys instead of numeric indexes.

The activity map in the spec was modified so that the keys were out of sort order.

[#170757052]

https://www.pivotaltracker.com/story/show/170757052

### DIFF
--- a/js/containers/dashboard/dashboard-app.js
+++ b/js/containers/dashboard/dashboard-app.js
@@ -15,7 +15,7 @@ import Header from "../../components/common/header";
 import HelpModal from "../../components/dashboard/help-modal";
 import DataFetchError from "../../components/report/data-fetch-error";
 import LoadingIcon from "../../components/report/loading-icon";
-import { getActivityTrees } from "../../selectors/report-tree";
+import { getActivityTrees, getSequenceTree } from "../../selectors/report-tree";
 import { getStudentProgress, getSortedStudents, getSelectedQuestion } from "../../selectors/dashboard-selectors";
 import css from "../../../css/dashboard/dashboard-app.less";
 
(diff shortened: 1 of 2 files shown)


knowuh:
FWIW, I tested this report with the offering discussed in the PT story:

https://learn.concord.org/api/v1/offerings/49907

And it sorts that dashboard appropriately.

js/containers/dashboard/dashboard-app.js
```diff
@@ -51,7 +51,11 @@ class DashboardApp extends PureComponent {
 
   render() {
     const { initialLoading } = this.state;
-    const { error, clazzName, activityTrees, students, studentProgress, expandedStudents, expandedActivities, expandedQuestions, setActivityExpanded, setStudentExpanded, setQuestionExpanded, setStudentsExpanded, setStudentSort, selectedQuestion, selectQuestion, trackEvent } = this.props;
+    const { error, clazzName, sequenceTree, students, studentProgress, expandedStudents, expandedActivities, expandedQuestions, setActivityExpanded, setStudentExpanded, setQuestionExpanded, setStudentsExpanded, setStudentSort, selectedQuestion, selectQuestion, trackEvent } = this.props;
+
+    // In order to get list the activies in correct right order,
```
kswenson:
```
    // In order to list the activities in correct order,
```

PR #148, opened 2020-04-08
Add prop to interactive iframe to allow fullscreen

sfentress:
[#171961435]


sfentress:
`allow="..."` is the modern iframe style, testing locally found it worked fine.

PR #149, opened 2020-05-14
Update dependencies and project configuration

kswenson:
TL;DR
- update nearly all dependencies to latest (exceptions noted below)
- switch from deprecated TSLint to ESLint (required by some dependencies)
  - required small changes to a large number of files which explains the changed file count
- switch from `react-dom/test-utils` to [React Testing Library](https://testing-library.com/docs/react-testing-library/intro) for some tests (the former tests failed after the dependency updates)

Note: The feedback cypress test is flaky and seems to fail about 50% of the time.

Dependencies not updated to latest:
- `@cypress/code-coverage` updated to 2.0.5 but not to 3.7.2 as 3.7.2 requires some configuration changes (which were simple) and then broke some of our cypress tests in ways that weren't trivially obvious to fix so I punted.
- `aws-sdk` is currently at 2.457.0 and I chose not to update to 2.677.0 at this time as I'm not confident that our existing tests are sufficient to identify whether the update broke anything in our portal communication. Should be considered separately.
- `bootstrap` 3.4.1 not updated to 4.5 as there are a lot of breaking styling changes in 4.x and I'm not sure how best to make sure that nothing changed/broke in appearance. If we care about the update (we may not) it should be considered separately.
- `eslint` 6.8.0 installed rather than 7.0.0 as some dependencies have not yet been updated to work with 7.x.
- `immutable` 3.8.2 not updated to 4.0.0-rc.12 as the latter is an unfinished release candidate of an unreleased major version of an unsupported project. Seems safer to stick with what we have.
- `jest` updated to 25.5.4 but not to 26.0.1 as `ts-jest` is not yet compatible with the latest jest version.

@scytacki Consider yourself an optional reviewer. I tagged you mainly as an FYI.


scytacki:
The feedback tests seem to be doing a little better than 50%, too bad they've become flakey.

For context on them, they are using Firestore in an "offline" mode. First the tests pretends to be a teacher and provide feedback on the mock student data. This is then saved to firestores offline local storage.  Then the tests pretends to be a student viewing their work, and it should be loading in the feedback data stored by firestore offline. 

This approach was working well before.

js/api.ts
```diff
@@ -252,13 +252,12 @@ export function updateReportSettingsInPortal(data: any) {
...
     }).then(checkStatus);
   } else {
-    // tslint:disable-next-line:no-console
     console.warn("No OFFERING/REPORT_URL. Faking put method.");
     return new Promise(resolve => resolve({}));
   }
 }
 
-export function feedbackSettingsFirestorePath(sourceKey: string, instanceParams?: { platformId?: string, resourceLinkId?: string }) {
+export function feedbackSettingsFirestorePath(sourceKey: string, instanceParams?: { platformId?: string; resourceLinkId?: string }) {
```
scytacki, line 260: Is the `;` right here?

js/api.ts
```diff
@@ -133,8 +133,8 @@ export function fetchFirestoreJWT(classHash: string, firebaseApp?: string) {
 }
 
 export function authFirestore(rawFirestoreJWT: string) {
-  return signInWithToken(rawFirestoreJWT).catch(err => {
-    // tslint:disable-next-line no-console
+  const authResult = signInWithToken(rawFirestoreJWT) as Promise<any>;
```
pjanik, line 136: I think this should be Promise<firebase.auth.UserCredential | void> if we want to be fully correct, but it doesn't matter as we don't use return value anyway. I'm surprised it requires manual casting tho. Why did you have to change these lines and cast type? Was it ESLint thing?

js/api.ts
kswenson: It's an interface definition, not a JavaScript object so yes semicolon is correct, although it was not immediately obvious to me, either. It was found by the linter.

kswenson:
>I think it's appropriate to add changes to this PR to get the .less files working on Windows systems.

I agree that this seems like part of basic project configuration and so makes sense to add a commit to this PR. I had thought that the solution might involve use of `path.join` which handles platform-specific path separators properly, but I guess if we can get it to work by just extending the regex that seems reasonable as well, although we'll have to change the regex the next time we add support for another platform with an entirely different path separator. 😉

js/api.ts
kswenson: Newer version of TypeScript complained about incompatible types. I think they tightened up some of their type-checking around promises somewhere along the way. As I recall I tracked it down to TS 3.6. I'll try what you suggest because that would be a better workaround.

kswenson: FYI, without the cast the error is:
```
ERROR in /Users/kswenson/Development/cc-dev/portal-report/js/api.ts
ERROR in /Users/kswenson/Development/cc-dev/portal-report/js/api.ts(137,21):
TS2349: This expression is not callable.
  Each member of the union type '(<TResult = never>(onrejected?: ((reason: any) => TResult | PromiseLike<TResult>) | null | undefined) => Promise<void | TResult>) | (<TResult = never>(onrejected?: ((reason: any) => TResult | PromiseLike<...>) | ... 1 more ... | undefined) => Promise<...>)' has signatures, but none of those signatures are compatible with each other.
```

kswenson: Excellent! The cast to `Promise<firebase.auth.UserCredential | void>` you suggested worked so I'll make that change. I thought I had tried that variant but apparently not.

kswenson:
I played around with the flaky feedback cypress test. It always seems to work locally whether I run it with `npm run cypress:test` or `npm run cypress:open`. @scytacki is right that on Travis the failure rate is lower than the 50% I stated earlier. I think I had had two failures in the previous four builds at the time I wrote that but taking a larger sample size maybe it fails ~20% of the time, so I'm going to go ahead and merge this and we can keep an eye on it.

PR #153, opened 2020-05-20
Standalone iframe

sfentress:
This creates a stand-alone iframe container, to view full-size versions of student models, which can be viewed by including the url query parameters `iframeQuestionId` and `studentId`.

E.g. http://portal-report.concord.org/branch/172324236-standalone-iframe/?iframeQuestionId=mw_interactive_19&studentId=1

This also adds links for opening the stand-alone view to the main report, alongside the iframe toggles.

Currently, this just adds the link to all answers that have the property `displayInIframe`, i.e. those that should show embedded. Eventually we may want to have additional authorable properties defined in LARA which determine how the models should be displayed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,10 +127,13 @@ when you view the class details, you will see a "Local Tracked Q" button next to
 
 Besides the parameters needed for loading data, the report also supports
 
-* `studentId`:  This shows the report for a single student, and removes some UI affordances. The filtering of the student
-                data happens client-side.
-* `iframeQuestionId`: This, combined with a valid `studentId`, will show a stand-alone, full-size iframe containing the
-                model referenced by iframeQuestionId, and the answer saved by studentId (either as state or as a url).
+* `studentId={id}`:   This shows the report for a single student, and removes some UI affordances. The filtering of the
+                      student data happens client-side.
+* `iframeQuestionId={id}`: This, combined with a valid `studentId`, will show a stand-alone, full-size iframe containing
+                      the model referenced by iframeQuestionId, and the answer saved by studentId (either as state or as
+                      a url).
+* `enableFirestorePersistence=true`: Uses a local firestore DB for data persistance across sessions and tabs. Clear the
+                      DB by going to `dev tools > Application > IndexedDB > firebaseLocalStorageDb > Delete database`
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -129,6 +129,8 @@ Besides the parameters needed for loading data, the report also supports
 
 * `studentId`:  This shows the report for a single student, and removes some UI affordances. The filtering of the student
                 data happens client-side.
+* `iframeQuestionId`: This, combined with a valid `studentId`, will show a stand-alone, full-size iframe containing the
+                model referenced by iframeQuestionId, and the answer saved by studentId (either as state or as a url).
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -104,6 +104,32 @@ Additional, useful resources:
 
 Note that conventions in the dashboard part of the code base are somewhat different than in the report part.
 
+### Using data
+
+Data is fetched using `api.js`.
+
+If the query parameters of the url does not include values for `offering` and `class`, we will load in fake data from
+the `js/data` folder. This data gets loaded in much the same way as real data, so can be used for testing.
+
+If we do have `offering` and `class` parameters, then `api.js` will first attempt to get the data for the offering and
+class from the portal. To do this it also needs a `token` parameter, which is used to authenticate with the portal and
+expires after one hour. Besides the class and offering data, we will also fetch a firestore JWT from the portal, given
+the classHash (from the fetched class data) and the token. Using this JWT, we can authenticate with Firestore. Once we
+have successfully authenticated, `receivePortalData` is called in `index.ts`, which starts watching the sequence
+structure and answer data.
+
+To test the portal using real data, the easiest way is simply to open a report as a teachers from the portal, and then
+replace the url host and path with `localhost:8080`. Alternatively, if you are able to edit the portal settings for the
+offering, you can add the "Developers Tracked Questions (Local)" report to the External Reports of the offering, and
+when you view the class details, you will see a "Local Tracked Q" button next to "Report" which will link to localhost.
+
+### Additional URL Parameters
+
+Besides the parameters needed for loading data, the report also supports
+
+* `studentId`:  This shows the report for a single student, and removes some UI affordances. The filtering of the student
+                data happens client-side.
+
 ## License
 
 [MIT](https://github.com/concord-consortium/grasp-seasons/blob/master/LICENSE)
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ Note that conventions in the dashboard part of the code base are somewhat differ
 
 Data is fetched using `api.js`.
 
-If the query parameters of the url does not include values for `offering` and `class`, we will load in fake data from
+If the query parameters of the url do not include values for `offering` and `class`, we will load in fake data from
 the `js/data` folder. This data gets loaded in much the same way as real data, so can be used for testing.
 
 If we do have `offering` and `class` parameters, then `api.js` will first attempt to get the data for the offering and
@@ -118,7 +118,7 @@ the classHash (from the fetched class data) and the token. Using this JWT, we ca
 have successfully authenticated, `receivePortalData` is called in `index.ts`, which starts watching the sequence
 structure and answer data.
 
-To test the portal using real data, the easiest way is simply to open a report as a teachers from the portal, and then
+To test the portal using real data, the easiest way is simply to open a report as a teacher from the portal, and then
 replace the url host and path with `localhost:8080`. Alternatively, if you are able to edit the portal settings for the
 offering, you can add the "Developers Tracked Questions (Local)" report to the External Reports of the offering, and
 when you view the class details, you will see a "Local Tracked Q" button next to "Report" which will link to localhost.
--- a/js/components/report/data-fetch-error.js
+++ b/js/components/report/data-fetch-error.js
@@ -18,7 +18,7 @@ export default class DataFetchError extends PureComponent {
     </div>;
   }
 
-  renderGenericInfo(error) {
+  renderGenericUrlInfo(error) {
     return (
       <div>
         <div>URL: {error.url}</div>
@@ -28,7 +28,18 @@ export default class DataFetchError extends PureComponent {
     );
   }
 
+  renderGenericMessage(error) {
+    return (
+      <div>
+        <div>{error}</div>
+      </div>
+    );
+  }
+
   renderError(error) {
+    if (error.body) {
+      return this.renderGenericMessage(error.body);
+    }
     switch (error.status) {
       case 401:
         return this.renderUnauthorized();
@@ -37,7 +48,7 @@ export default class DataFetchError extends PureComponent {
       case 599:
         return this.renderNetworkError(error);
       default:
-        return this.renderGenericInfo(error);
+        return this.renderGenericUrlInfo(error);
     }
   }
 
@@ -45,7 +56,7 @@ export default class DataFetchError extends PureComponent {
     const { error } = this.props;
     return (
       <div className="data-fetch-error">
-        <h2>Connection to server failed</h2>
+        <h2>{error.title || "Connection to server failed"}</h2>
         {this.renderError(error)}
       </div>
     );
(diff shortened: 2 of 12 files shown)


js/containers/report/iframe-standalone-app.js
```diff
@@ -0,0 +1,104 @@
...
+  const answers = reportState && reportState.get("answers");
+  const dataDownloaded = !error && !data.get("isFetching");
+  return {
+    report: dataDownloaded && reportState,
+    iframeQuestionId: iframeQuestionId,
+    answers: answers,
+    isFetching: data.get("isFetching"),
+    error: error,
+  };
+}
```
kswenson, line 90: I'd use the shorthand syntax here:
```javascript
  return {
    report: dataDownloaded && reportState,
    iframeQuestionId,
    answers,
    isFetching: data.get("isFetching"),
    error,
  };
```
We could also consider enabling the [object-shorthand](https://eslint.org/docs/rules/object-shorthand) lint rule, although it would require (auto-)fixing existing instances. Thoughts?

js/components/report/interactive-iframe.js
```diff
@@ -42,14 +42,14 @@ export default class InteractiveIframe extends PureComponent {
...
-    const { src, width, height } = this.props;
+    const { src, width, height, style } = this.props;
     return (
       // eslint-disable-next-line react/no-string-refs
       <iframe ref="iframe"
         src={src}
         width={width || "300px"}
         height={height || "300px"}
-        style={{border: "none", marginTop: "0.5em"}}
+        style={style || {border: "none", marginTop: "0.5em"}}
```
kswenson, line 52: Is there a scenario in which merging styles would make more sense than choosing one or the other?

js/containers/app.tsx
```diff
@@ -20,13 +21,14 @@ interface IProps {
...
-    return (
-      viewType === PORTAL_DASHBOARD
-        ? <PortalDashboardApp />
-        : viewType === DASHBOARD
-          ? <DashboardApp />
-          : <ReportApp />
-    );
+    switch (viewType) {
+      case IFRAME_STANDALONE:
+        return <IframeStandaloneApp />;
```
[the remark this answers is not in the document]
sfentress: merge-error. Will correct.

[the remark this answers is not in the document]
sfentress: Thanks.

js/containers/report/iframe-standalone-app.js
sfentress: We could put that in, but I'd do that as a separate linting PR, so as not to muddy waters.

js/components/report/interactive-iframe.js
sfentress: I debated between them, but the specific case was wanting to not use a default style (`marginTop: 0.5em`). So if we merged we'd have to pass in an additional clearing style (`marginTop: 0`), instead of just omitting it.

While in this case passing in an additional prop isn't much, I think it's additional confusing/source for errors if you pass in a bunch of props, and get additional props you didn't expect.

js/components/report/iframe-answer.js
```diff
@@ -36,22 +32,47 @@ export default class IframeAnswer extends PureComponent {
...
     } catch (e) {
       return answer;
     }
   }
 
+  /**
+   * Adds a studentId and iframeQuestionId to the existing url
+   */
+  getStandaloneLinkUrl(question, answer) {
+    const baseUrl = `${window.location.origin}${window.location.pathname}`;
```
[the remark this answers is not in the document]
sfentress: I'm not sure what the current use-case would be. Unless we start pushing state into the browser history as the teacher clicks around. But then we wouldn't necessarily want that maintained when viewing the standalone iframe.

It is possible, it's true, but there would already be a number of places that url location code would need to be updated once we did that, and we'd probably need to consider each individually instead of being surprised.

js/components/report/iframe-answer.js
```diff
@@ -36,22 +32,47 @@ export default class IframeAnswer extends PureComponent {
...
-    const decorator = question.get("displayInIframe") ? "" : <span className="pr-icon-external-link" />;
-    return <a href={linkUrl} onClick={this.toggleIframe} target="_blank">View Work {decorator}</a>;
+    const externalLinkIcon = <span className="pr-icon-external-link" />;
+    if (question.get("displayInIframe")) {
+      const toggleText = iframeVisible ? "Hide" : "View Work";
+      const standaloneLinkUrl = this.getStandaloneLinkUrl(question, answer);
+      return <React.Fragment>
+      <a onClick={this.toggleIframe} target="_blank">{toggleText}</a> |{" "}
+      <a href={standaloneLinkUrl} target="_blank">Open in new tab {externalLinkIcon}</a>
+    </React.Fragment>;
```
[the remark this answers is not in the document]
sfentress: :+1:

README.md
```diff
@@ -104,6 +104,34 @@ Additional, useful resources:
...
+when you view the class details, you will see a "Local Tracked Q" button next to "Report" which will link to localhost.
+
+### Additional URL Parameters
+
+Besides the parameters needed for loading data, the report also supports
+
+* `studentId`:  This shows the report for a single student, and removes some UI affordances. The filtering of the student
+                data happens client-side.
+* `iframeQuestionId`: This, combined with a valid `studentId`, will show a stand-alone, full-size iframe containing the
+                model referenced by iframeQuestionId, and the answer saved by studentId (either as state or as a url).
```
scytacki, line 133: There is another optional url parameter here that is used by the tests to tell the report to run firestore offline and not make a connection.

js/components/report/iframe-answer.js
```diff
@@ -12,13 +13,8 @@ export default class IframeAnswer extends PureComponent {
...
 
-  toggleIframe(event) {
-    const { question } = this.props;
-    if (question.get("displayInIframe")) {
-      // If displayInIframe == true, we won't follow the link.
-      event.preventDefault();
-      this.setState({iframeVisible: !this.state.iframeVisible});
-    }
+  toggleIframe() {
+    this.setState({iframeVisible: !this.state.iframeVisible});
```
scytacki, line 17: This line is untested. It be great to add a cypress test to toggle an iframe.

js/components/report/iframe-answer.js
```diff
@@ -36,22 +32,49 @@ export default class IframeAnswer extends PureComponent {
     */
     try {
       const json = JSON.parse(answer);
-      const interactiveState = JSON.parse(json.interactiveState);
-      return interactiveState.lara_options.reporting_url;
+      const interactiveState = json.interactiveState ? JSON.parse(json.interactiveState) : null;
+      if (interactiveState && interactiveState.lara_options && interactiveState.lara_options.reporting_url) {
+        return interactiveState.lara_options.reporting_url;
```
scytacki, line 37: This line is also untested. A cypress test of this would also be great.

js/components/report/iframe-answer.js
```diff
@@ -12,13 +13,8 @@ export default class IframeAnswer extends PureComponent {
     this.toggleIframe = this.toggleIframe.bind(this);
   }
 
-  toggleIframe(event) {
-    const { question } = this.props;
-    if (question.get("displayInIframe")) {
-      // If displayInIframe == true, we won't follow the link.
-      event.preventDefault();
-      this.setState({iframeVisible: !this.state.iframeVisible});
-    }
```
scytacki, line 21: I'm pretty certain that LabBooks have a reporting_url, but they should be shown in an iframe. I remember we talked about this, so probably this is now handled elsewhere. I just wanted to flag it, incase it was missed.

js/components/report/iframe-answer.js
```diff
@@ -36,22 +32,49 @@ export default class IframeAnswer extends PureComponent {
...
+        </React.Fragment>
+      );
+    } else {
+      return <a href={linkUrl} target="_blank">View work in new tab {externalLinkIcon}</a>;
+    }
   }
 
   renderIframe() {
-    const { answer, question, alwaysOpen } = this.props;
+    const { answer, question } = this.props;
```
scytacki, line 77: Apparently this is untested too. I guess the cypress tests are not toggling iframes which I think would explain why this is untested. Adding that test would be great.

js/components/report/data-fetch-error.js
```diff
@@ -28,7 +28,18 @@ export default class DataFetchError extends PureComponent {
...
+    return (
+      <div>
+        <div>{error}</div>
+      </div>
+    );
+  }
+
   renderError(error) {
+    if (error.body) {
+      return this.renderGenericMessage(error.body);
```
scytacki, line 41: This is also untested, but I suspect testing this will be harder. So it'd be lower priority than the other untested lines.

js/containers/app.tsx
```diff
@@ -12,13 +13,17 @@ interface IProps {
...
-          ? <DashboardApp />
-          : <ReportApp />
-    );
+    switch (viewType) {
+      case DASHBOARD:
+        return <DashboardApp />;
+      case PORTAL_DASHBOARD:
+        return <PortalDashboardApp />;
+      case IFRAME_STANDALONE:
+        return <IframeStandaloneApp />;
```
scytacki, line 22: This is untested, so I'd guess there isn't a basic cypress test to open the iframe standalone.

js/containers/report/iframe-standalone-app.js
```diff
@@ -0,0 +1,104 @@
...
+import LoadingIcon from "../../components/report/loading-icon";
+import InteractiveIframe from "../../components/report/interactive-iframe";
+
+import "../../../css/report/report-app.less";
+import "../../../css/report/iframe-standalone-app.less";
+import config from "../../config";
+
+class IframeStandaloneApp extends PureComponent {
+  constructor(props) {
+    super(props);
```
scytacki, line 17: Lots of untested code in this file.

js/containers/report/iframe-standalone-app.js
```diff
@@ -0,0 +1,104 @@
...
+      state = null;
+    } else if (answer.get("type") === "interactive_state") {
+      // URL field is provided by question. Answer field is a state that will be passed
+      // to the iframe using iframe-phone.
+      url = question.get("url");
+      state = answer.get("answer");
+    }
+    return (
+      <InteractiveIframe src={url} state={state} style={{border: "none"}} width="100%" height="100%" />
+    );
```
scytacki, line 59: With the exception of the width and height this is identical to: 
https://github.com/concord-consortium/portal-report/blob/172324236-standalone-iframe/js/components/report/iframe-answer.js#L78-L95

js/containers/report/iframe-standalone-app.js
```diff
@@ -0,0 +1,104 @@
...
+  return {
+    fetchAndObserveData: () => dispatch(fetchAndObserveData()),
+    hideUnselectedQuestions: () => dispatch(hideUnselectedQuestions()),
+    showUnselectedQuestions: () => dispatch(showUnselectedQuestions()),
+    setNowShowing: (nowShowingValue, selectedStudentIds) => dispatch(setNowShowing(nowShowingValue, selectedStudentIds)),
+    setAnonymous: value => dispatch(setAnonymous(value)),
+    hideCompareView: () => dispatch(hideCompareView()),
+    trackEvent: (category, action, label) => dispatch(trackEvent(category, action, label)),
+  };
+};
```
scytacki, line 102: Not sure how easy this is to refactor but it looks identical to: https://github.com/concord-consortium/portal-report/blob/172324236-standalone-iframe/js/containers/report/report-app.js#L95

scytacki:
@sfentress I mentioned this as the beginning of the interactives meeting and sorry if I'm repeating myself: If you install the codeclimate chrome extension it shows you the places where code is not covered by tests. It also shows the issues that codeclimate finds like redundant code and complex functions.

js/components/report/iframe-answer.js
sfentress: Yup. LabBooks have `displayInIframe`, which is respected. The refactoring removed a useless href in the toggle link for embedded iframes, and only calls `toggleIframe` for embedded iframes (including LabBooks), so this check was no longer needed.

sfentress:
Sounds good. I had been ignoring the Code Climate checks because they were so noisy, but will add the tests, which will bump the code coverage. (And actually I hadn't heard of the Chrome extension, so will install that, thanks.)

README.md
sfentress: :+1:

scytacki:
This looks good to me. There is still some duplicate code, but that seems OK for now.
 The feedback tests still have problems, but that should be a different PR.

PR #154, opened 2020-05-22
Add mock data

eireland:
Added a class of 36 students with variety of names
Added an activity with variety of question types, interactive types, and correct and wrong answers

### DIFF
--- a/cypress/integration/geode-dashboard-smoke-test.spec.js
+++ b/cypress/integration/geode-dashboard-smoke-test.spec.js
@@ -14,7 +14,7 @@ context("Geode Dashboard Smoke Test", () => {
 
     beforeEach(() => {
         cy.visit("/?dashboard=true");
-        cy.fixture("class-data.json").as("classData");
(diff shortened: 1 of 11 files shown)


sfentress:
@eireland The data looks fine, but it's a little confusing to me how this should be used. I think it should at least be documented, and maybe even a url parameter or something could be added.

I worked out that I could see the data if I swapped the commented-out line importing `fakeAnswers`, and swapped the line importing `fakeActivityStructure` and also renamed it `fakeSequenceStructure`.

Was that the intent, or if this supposed to eventually replace all the data? In which case, should that be a part of the PR?

Also, there are new students for testing things like very long names, such as "Hubert Wolfe­schlegel­stein­hausen­berger­dorff," but as far as I can tell these students don't appear in the report?

eireland:
For now, the only way to see this mock data is by hard coding it into the actions/index.ts codeline. It can be used later when class switching and activity/sequence switching are available. I can look into adding query params for now.

PR #156, opened 2020-05-22
Add progress legend

eireland:
Added layout of progress legend. Will still need the correct font family to be specified.


sfentress:
This is looking good, but I think there are a few things that can be tweaked to get this closer to Michael's spec (https://app.zeplin.io/project/5e96f4b8527afa7d28bb1468/screen/5ec52a802111b03333e5583d):

* Legend not `position: fixed` (that causes it not to scroll)
* 19px from the bottom of the container. The easiest way to do that right now would be to making the wrapping container shorter (I think `height: 205px`), and then positioning the legend at the bottom by adding `position: relative` to the wrapper and `position: absolute; bottom: -19px` to the legend.
* border 1.5px, on the legend and the bottom border that makes the line
* `box-sizing: border-box;` This will make the actual size 32px high, as it will take into account the size of the borders and padding. This will then require some futzing to get the text to line up in the middle
* Consider using css for icons, I could work out the css for them if you like.

cypress/integration/portal-dashboard-smoke-test.spec.js
```diff
@@ -0,0 +1,15 @@
...
+    describe('progress',()=>{
+        it('verify progress legend loads',()=>{
+            const progress=["Completed","In progress", "Not yet started"];
+            cy.get('[data-test=progress-legend]').should('be.visible');
+            progress.forEach((prog)=>{
+                cy.get('[data-test="'+prog+'-legend"]').should('be.visible');
+            });
+        });
+    });
+});
```
mklewandowski, line 15: Could we add an extra newline here?  Kirk pushed me to be consistent with doing this.  Also thanks for adding this model.  I'm going to reuse it to add tests for the other components.

js/components/portal-dashboard/legend-container.tsx
```diff
@@ -0,0 +1,41 @@
...
+            <div key={index} className={css.legendKey} data-test={progress.name + "-legend"}>
+              <div className={progress.class}></div>
+              <div className={css.legendText}>{progress.name}</div>
+            </div>
+          );
+        })}
+      </div>
+    );
+  }
+}
```
mklewandowski, line 41: Add a newline at the end of this file.  See above.

js/components/portal-dashboard/level-viewer.tsx
```diff
@@ -1,12 +1,14 @@
...
 
 import css from "../../../css/portal-dashboard/level-viewer.less";
+import { ProgressLegendContainer } from "./legend-container";
 
 export class LevelViewer extends React.PureComponent {
   render() {
     return (
       <div className={css.levelViewer}>
         level viewer content
+        <ProgressLegendContainer></ProgressLegendContainer>
```
mklewandowski, line 11: Write this using shorthand
`<ProgressLegendContainer />`

css/portal-dashboard/level-viewer.less
```diff
@@ -1,5 +1,9 @@
 @import "../variables";
 
 .levelViewer {
-  height: 240px;
+  height: 205px;
+  width: 100%;
```
mklewandowski, line 5: The 100% width does cause the left class nav component size to get messed up.  I'll see if there's an obvious solution.

css/portal-dashboard/progress-legend.less
```diff
@@ -0,0 +1,77 @@
...
+    line-height: 17px;
+    vertical-align: text-bottom;
+    position: relative;
+  }
+  .completed {
+    margin: 5px;
+    position:relative;
+    bottom:3px;
+    box-sizing: border-box;
+    border: 1px solid #ff8415;
```
mklewandowski, line 45: Colors like this can now be pulled out of the new `variables.less` if you rebase onto master.  So this color would be `@cc-orange`.  Just make sure to import the following in the top of the .less file: `@import "./variables";`

css/portal-dashboard/level-viewer.less
mklewandowski: This works as an alternative so the class nav still behaves the same:
`width: calc(100vw - 240px);`

mklewandowski: But long term we might need to think about how these components slot together.

PR #160, opened 2020-05-27
172806194 student list

mklewandowski:
This PR adds small changes to the student list component to conform with the UI Spec.  Changes include:
- show ellipsis for long names
- add `isAnonymous` to props which will be used in a future story when implementing the slider to make student names anonymous
- make content scroll vertically if it overflows
- add a compact CSS class to handle future story where user can display rows in compact mode with a smaller height


css/portal-dashboard/student-list.less
```diff
@@ -3,4 +3,6 @@
 .studentList {
   width: 100%;
   background-color: white;
+  overflow: auto;
+  height: calc(100vh - 245px - 60px);
```
kswenson, line 7: Seems like a good place for some variables rather than magic numbers.

mklewandowski: Yeah, I was waiting for those components to settle down, but now is as good of a time as any to slot in some variables.

mklewandowski: Fixed.  Thanks!

PR #161, opened 2020-05-28
172806422 sort student list

mklewandowski:
Draft PR that adds a custom dropdown select menu and uses it to set the student sorting.

### DIFF
--- a/js/components/portal-dashboard/custom-select.tsx
+++ b/js/components/portal-dashboard/custom-select.tsx
@@ -6,6 +6,7 @@ interface IProps {
   items: SelectItem[];
   setStudentSort: any;
   trackEvent: any;
+  iconId: string;
 }
 
 interface IState {
@@ -20,7 +21,6 @@ export interface SelectItem {
 
 export class CustomSelect extends React.PureComponent<IProps, IState> {
   private divRef = React.createRef<HTMLDivElement>();
-  // private setDivRef: (element: any) => void;
   private setDivRef = (element: any) => {
     this.divRef = element;
   };
@@ -31,10 +31,6 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
       current: props.items[0].action,
       showList: false
     };
-
-    //this.setDivRef = (element) => {
-    //  this.divRef = element;
-    //};
   }
 
   public componentDidMount() {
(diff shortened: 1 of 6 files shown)


js/containers/portal-dashboard/portal-dashboard-app.tsx
```diff
@@ -91,7 +93,7 @@ function mapStateToProps(state: RootState) {
   };
 }
 
-const mapDispatchToProps = (dispatch: any, ownProps: any) => {
+const mapDispatchToProps = (dispatch: any, ownProps: any): Partial<IProps> => {
```
sfentress, line 96: The `Partial<IProps>` return type could also potentially be added to `mapStateToProps`, which exposes a very minor Typescript complaint about the type of `clazzName`. Not that important at all, but might be nice to add just as an example of writing a complete container component in TS, since we all like to copy from what has already been done elsewhere.

js/components/portal-dashboard/custom-select.tsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  private handleListClick = (current: string) => () => {
+    this.props.setStudentSort(current);
+    this.props.trackEvent("Dashboard", "Sorted", current);
+    this.setState({
+      current,
+      showList: false
+    });
+  }
```
sfentress, line 97: I assume it would eventually make sense for `<CustomSelect>` to just be passed an `onSelect` by its parent, so that it can be used for more than just sorting students. It doesn't need to be part of this PR.

js/containers/portal-dashboard/portal-dashboard-app.tsx
mklewandowski: Good thinking.  I made this change and added additional fixes as needed.

js/components/portal-dashboard/custom-select.tsx
mklewandowski: I changed it so it takes `onSelectItem`.  Should get us closer to generalizing this component.

js/containers/portal-dashboard/portal-dashboard-app.tsx
sfentress: Looks great. The other way to do it, if you didn't want the prop `clazzName` to be `string | false` would have been

```
function mapStateToProps(state: RootState): Partial<IProps> {
 ...
  return {
    isFetching: data.get("isFetching"),
    ...,
    clazzName: dataDownloaded ? state.getIn(["report", "clazzName"]) : undefined,
  };
}
```
but this is also totally fine.

mklewandowski: Yeah, I like this better.  I'm not crazy about or-ing in that boolean in the props.  Feels confusing.  Fixed.

PR #163, opened 2020-05-29
Add component Cypress tests

mklewandowski:
This PR adds a handful of simple Cypress tests to the top level portal dashboard components.

### DIFF
--- a/js/components/portal-dashboard/class-nav.tsx
+++ b/js/components/portal-dashboard/class-nav.tsx
@@ -17,13 +17,14 @@ export class ClassNav extends React.PureComponent<IProps> {
                                  { action: SORT_BY_LEAST_PROGRESS, name: "Least Progress" }];
     const { clazzName, setStudentSort, trackEvent } = this.props;
     return (
-      <div className={css.classNav}>
+      <div className={css.classNav} data-test="class-nav">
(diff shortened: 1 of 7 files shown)


js/components/portal-dashboard/class-nav.tsx
```diff
@@ -17,13 +17,14 @@ export class ClassNav extends React.PureComponent<IProps> {
                                  { action: SORT_BY_LEAST_PROGRESS, name: "Least Progress" }];
     const { clazzName, setStudentSort, trackEvent } = this.props;
     return (
-      <div className={css.classNav}>
+      <div className={css.classNav} data-test="class-nav">
```
sfentress, line 20: Is there a difference in intent between doing `data-test` and `data-cy`? I see the latter is used throughout the repo, and the former in only a couple spots.

mklewandowski: The Cypress best-practices docs don't specify a preference between the two.  However, for consistency, I think we should stick with what the project was doing.  Since previously the project only had 2 instances of `data-test` and over 100 instances of `data-cy`, I think we should convert these new tests to use `data-cy` and continue using that moving forward.  @eireland do you agree?

PR #164, opened 2020-05-29
Add number of students

eireland:
adds the student count to class nav


css/portal-dashboard/num-students.less
```diff
@@ -0,0 +1,23 @@
...
+.numStudentsContainer {
+    width: 168px;
+    height: 30px;
+    border-radius: 15px 0px 0px 15px;
+    background-color: @cc-teal-light7;
+    position: absolute;
+    top: 210px;
+    left: 72px;
+    font-size: 16px;
+    font-family: lato, arial, helvetica, sans-serif;
```
mklewandowski, line 12: I think you can remove this now.  I added the font-family to the top level component so all the children will get it too.

css/portal-dashboard/num-students.less
```diff
@@ -0,0 +1,23 @@
...
+    color: @cc-charcoal;
+}
+.containerLabel {
+    font-weight: bold;
+    padding-left:20px;
+}
+.numStudentsContainer span {
+    position: relative;
+    top: 6px;
+}
```
mklewandowski, line 23: Add newline at end of file.  We should add this to our linter (we tend to in other projects).

cypress/integration/portal-dashboard-smoke-test.spec.js
```diff
@@ -12,4 +12,10 @@ context("Portal Dashboard UI",()=>{
             });
         });
     });
+    describe('Number of students in class',()=>{
+        let numStudents = 6; //TODO: it would be better if numStudents is not hard coded.
+        it('verify number of students in a class loads',()=>{
+            cy.get('[data-test=num-student-container]').should('be.visible').and('contain', numStudents+ ' students');
```
mklewandowski, line 18: Sam mentioned this in my PR so I'll ask the same question.  Most of the project uses `data-cy` instead of `data-test`.  Do you think we should use `data-cy` to be consistent?

js/components/portal-dashboard/num-students-container.tsx
```diff
@@ -0,0 +1,19 @@
...
+    render() {
+      return (
+        <div className={css.numStudentsContainer} data-test="num-student-container">
+          <span className={css.containerLabel}>Class: </span>
+          <span data-test="student-number">{this.props.studentCount}</span>
+          <span> students</span>
+        </div>
+      );
+    }
+  }
```
mklewandowski, line 19: Add newline at end of file.

package.json
```diff
@@ -115,6 +115,7 @@
     "aws-sdk": "^2.451.0",
     "bootstrap": "^3.3.6",
     "chroma-js": "^2.1.0",
+    "eslint-plugin-cypress": "^2.10.3",
```
mklewandowski, line 118: Was this intended to be in this PR?

js/components/portal-dashboard/num-students-container.tsx
```diff
@@ -0,0 +1,19 @@
...
+  interface IProps {
+    studentCount: number;
+  }
+
+export class NumberOfStudentsContainer extends React.PureComponent<IProps> {
+    render() {
+      return (
+        <div className={css.numStudentsContainer} data-test="num-student-container">
+          <span className={css.containerLabel}>Class: </span>
+          <span data-test="student-number">{this.props.studentCount}</span>
```
mklewandowski, line 14: Same comment as above, should we use `data-cy` to be consistent with the rest of the project?

css/portal-dashboard/num-students.less
kswenson: I already added it, but I don't think our `.less` files get linted currently.

css/portal-dashboard/num-students.less
```diff
@@ -0,0 +1,23 @@
...
+    letter-spacing: 0;
+    color: @cc-charcoal;
+}
+.containerLabel {
+    font-weight: bold;
+    padding-left:20px;
+}
+.numStudentsContainer span {
+    position: relative;
+    top: 6px;
```
mklewandowski, line 22: I like to use flex boxes to center thing vertically and horizontally.  Here you might try just setting `line-height: 30px;` instead of `top: 6px` and see if it looks like the text is in the true vertical center.  On my system/screen that looks a pixel or so better than setting the top position.  Not sure if you see the same improvement.

js/components/portal-dashboard/num-students-container.tsx
kswenson: This should get caught by linter. One thing we could do is add a git hook so you can't push without a clean lint, for instance.

package.json
kswenson: As before, this should be in `devDependencies` not `dependencies`. As a result of this change it's now in both.

css/portal-dashboard/num-students.less
mklewandowski: Ah, good point.  For now, we should be careful to do this by hand if the linter doesn't catch it.

kswenson: Looks like [lesshint](https://www.npmjs.com/package/lesshint) supports this as well as some other potentially useful checks.

css/portal-dashboard/num-students.less
```diff
@@ -0,0 +1,23 @@
...
+    top: 210px;
+    left: 72px;
+    font-size: 16px;
+    font-family: lato, arial, helvetica, sans-serif;
+    letter-spacing: 0;
+    color: @cc-charcoal;
+}
+.containerLabel {
+    font-weight: bold;
+    padding-left:20px;
```
mklewandowski, line 18: Tiny issue.  Could we add a space after `padding-left`?  `padding-left: 20px;`

cypress/integration/portal-dashboard-smoke-test.spec.js
eireland: Yes! I got so used to putting data-test, forget this one uses this one used data-cy

package.json
eireland: Removed

cypress/integration/portal-dashboard-smoke-test.spec.js
mklewandowski: I'm going to change them to `data-cy` in my PR as well then.

PR #168, opened 2020-06-02
Add stub for class selection control

mklewandowski:
This PR adds in a UI stub for the class selection dropdown in the upper left navigation area.  A custom dropdown control is added that, at present, contains only the current, default class.  Eventually we will need to add other additional classes to the dropdown and change classes when a list item is selected.  However, this PR at least puts the basic UI in place.
image

### DIFF
--- a/css/portal-dashboard/class-nav.less
+++ b/css/portal-dashboard/class-nav.less
@@ -7,8 +7,8 @@
   width: 240px;
   background-color: @cc-teal-light6;
 
-  .title {
-    height: 65px;
+  .chooseClass {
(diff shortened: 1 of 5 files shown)


mklewandowski:
@sfentress let's see what Kiley thinks.  She had recommended putting in the feedback control even though it's functionality isn't working yet.  We can easily put the drop down in a disabled state so it serves as a mechanism to just show the class name for the time being.

mklewandowski:
Based on the demo/planning discussion, I'm going to merge this.

PR #170, opened 2020-06-02
173023374 response table

mklewandowski:
This PR adds a student response table that displays student activity progress as overall progress and as progress on specific activity answers.

A few notes:

- I didn't attempt to implement an animation when we transition from the response table summary to specific questions.  I also didn't attempt to implement having the level viewer scroll when the response table scrolls.  Those felt like they should be handled in a different PR once everything is in because there is a good chance that it might require restructuring divs to make both work.

- When we get an activity question, we look at the `type` to determine the question type and then display the proper UI for that question (for example, each question type has a unique icon).  However, I'm not seeing types for interactive table, interactive codap, image open response, or fill in the blank.  Is there more that we have to read from the activity question to parse out more specific sub-types for a given question?

- There are several `any` types could probably be cleaned up.

- There are still some subtle border issues when expanding content, but I'm not terribly concerned about those.


mklewandowski:
@sfentress Good feedback!  I made one final tweak to the scrolling, but will look into this in more depth in a later story.  I also updated the icons from today's discussion, but realize they will need to evolve as we get deeper.  Most importantly I added Cypress tests to confirm that we're displaying the progress and answer icons as expected.

PR #182, opened 2020-06-15
Support "answerText", listen to "height" iframe-phone message

pjanik:
Support new "answerText" metadata and "height" iframe-phone message.

### DIFF
--- a/__mocks__/iframe-phone.ts
+++ b/__mocks__/iframe-phone.ts
@@ -1,0 +1,27 @@
+const parentInstances: ParentEndpoint[] = [];
+
+class ParentEndpoint {
+  constructor(iframe: HTMLIFrameElement, phoneAnsweredCallback: () => void) {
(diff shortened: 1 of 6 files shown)


scytacki:
I finally figured out how to find this code coverage (it is no longer available in the PRs). So I found it by looking at the individual files. 

The missing coverage is on the handleHeight method.

pjanik:
@scytacki, I ended up adding Jest test instead of Cypress (not sure if you meant Cypress or any kind of coverage). Cypress could be useful, but we'd have to create some test interactive within this repo that actually sends these iframe-phone messages, add Cypress iframe helpers, etc etc. I think it'd be very nice, but might require a bit more time and better planning to make it useful later too.

I still had to mock iframe-phone in Jest, but this was a bit simpler and should be useful too. I've added a test that checks height message and extended "firebaseJWT" test that wasn't checking this piece.

PR #183, opened 2020-06-15
Add popup class nav

eireland:
Adds the class nav area with the view list by toggle and the spotlight toggle

### DIFF
--- a/css/portal-dashboard/all-responses-popup/popup-class-nav.less
+++ b/css/portal-dashboard/all-responses-popup/popup-class-nav.less
@@ -2,6 +2,7 @@
 
 .popupClassNav {
   width: 240px;
+  font-weight: normal;
 
   .icon {
     width: 20px;
@@ -21,6 +22,7 @@
   .viewListOption {
     background-color: @cc-teal-light5;
     border-bottom: 1px solid @cc-teal;
+    font-weight: bold;
 
     .optionIcon{
       width: 24px;
--- a/css/portal-dashboard/all-responses-popup/popup-class-nav.less
+++ b/css/portal-dashboard/all-responses-popup/popup-class-nav.less
@@ -3,111 +3,110 @@
 .popupClassNav {
   width: 240px;
 
-    .icon {
-        width: 20px;
-        height: 20px;
-        object-fit: contain;
-        fill: white;
-        margin: 0 5px 0 5px;
+  .icon {
+    width: 20px;
+    height: 20px;
+    object-fit: contain;
+    fill: white;
+    margin: 0 5px 0 5px;
+  }
+  .columnHeader{
+    display: flex;
+    flex-direction: row;
+    justify-content: center;
+    align-items: center;
+    height: 50px;
+    color: @cc-charcoal;
+  }
+  .viewListOption {
+    background-color: @cc-teal-light5;
+    border-bottom: 1px solid @cc-teal;
+
+    .optionIcon{
+      width: 24px;
+      height: 24px;
+      margin: 0 auto 0;
     }
-    .columnHeader{
-        display: flex;
-        flex-direction: row;
-        justify-content: center;
-        align-items: center;
-        height: 50px;
+    .toggle {
+      display: flex;
+      justify-content: center;
+      align-items: center;
+      width: 47px;
+      height: 30px;
+      background-color: @cc-teal-light5;
+      fill: @cc-teal;
+      border: 1.5px solid @cc-charcoal-light1;
+      box-sizing: border-box;
+      &:hover {
+        background-color: @cc-teal-light3;
+      }
+    }
+    .listByStudents {
+      border-radius: 15px 0 0 15px;
+      margin-right: 5px;
+      margin-left: 11px;
+
+      .optionIcon {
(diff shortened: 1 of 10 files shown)


eireland:
@mklewandowski I think the is ready for review now

css/portal-dashboard/all-responses-popup/popup-class-nav.less
```diff
@@ -0,0 +1,113 @@
+@import "../variables";
+
+.popupClassNav {
+  width: 240px;
```
mklewandowski, line 4: I think some bold fonts sneaked in.  Add this to a font-weight normal to fix in the popupClassNav class:
`font-weight: normal;`

js/components/portal-dashboard/all-responses-popup/popup-class-nav.tsx
```diff
@@ -0,0 +1,137 @@
...
+  }
+  private renderViewListOptions() {
+    let listByStudentClasses: string = css.toggle + " " + css.listByStudents,
+      listByQuestionsClasses: string = css.toggle + " " + css.listByQuestions;
+
+    if (!this.state.inQuestionMode) {
+      listByStudentClasses += "  " + css.selected;
+    } else {
+      listByQuestionsClasses += " " + css.selected;
+    }
```
mklewandowski, line 78: You could convert this chunk to something slightly smaller with ternary:
```
    !this.state.inQuestionMode
      ? listByStudentClasses += " " + css.selected
      : listByQuestionsClasses += " " + css.selected;
```

mklewandowski: And if ambitious, you could probably add it to the let statements above to shrink everything down:
```
    const listByStudentClasses: string = `${css.toggle} ${css.listByStudents} `
        + (!this.state.inQuestionMode ? css.selected : "");
    const listByQuestionsClasses: string = `${css.toggle} ${css.listByQuestions} `
        + (this.state.inQuestionMode ? css.selected : "");
```

That would replace the let and the if statements.

js/components/portal-dashboard/all-responses-popup/popup-class-nav.tsx
```diff
@@ -0,0 +1,137 @@
...
+  private handleStudentsToggleClick = () => {
+    this.setState({
+      inQuestionMode: false
+    });
+  }
+  private handleQuestionsToggleClick = () => {
+    this.setState({
+      inQuestionMode: true
+    });
+  }
```
mklewandowski, line 120: handleStudentsToggleClick and handleQuestionsToggleClick could be refactored into a single function that takes in a boolean, but isn't necessary.

js/components/portal-dashboard/all-responses-popup/popup-class-nav.tsx
```diff
@@ -0,0 +1,137 @@
...
+  private handleSpotlightClick = () => {
+    const wasSpotlightOn = this.state.isSpotlightOn;
+    this.setState({
+      isSpotlightOn: !wasSpotlightOn
+    });
```
mklewandowski, line 125: You can also use `prevState`:
```
this.setState(prevState => ({
  isSpotlightOn: !prevState.isSpotlightOn
}));
```

js/components/portal-dashboard/all-responses-popup/popup-class-nav.tsx
```diff
@@ -0,0 +1,137 @@
...
+    if (!wasSpotlightOn && !this.state.studentSelected) {
+      this.setState({ showDialog: true });
+    }
+  }
+
+  private closeShowDialog = (show: boolean) => {
+    this.setState({ showDialog: show });
+    this.setState({
+      isSpotlightOn: !this.state.isSpotlightOn
+    });
```
mklewandowski, line 135: You can condense these setStates into a single call:
```
    this.setState({
      isSpotlightOn: !this.state.isSpotlightOn,
      showDialog: show
    });
```

css/portal-dashboard/all-responses-popup/popup-class-nav.less
eireland: Good catch! I've been assuming all these labels are bold.

js/components/portal-dashboard/all-responses-popup/popup-class-nav.tsx
mklewandowski: Something like this:
```
  private setQuestionMode = (value: boolean) => () => {
    this.setState({
      inQuestionMode: value
    });
  }
```

And then, you can simply add JSX:
```
<div className={listByStudentClasses} data-cy="list-by-student-toggle" onClick={this.setQuestionMode(false)}>

<div className={listByQuestionsClasses} data-cy="list-by-questions-toggle" onClick={this.setQuestionMode(true)}>
```

eireland:
Made the suggested changes. cleaned up the css logic in a few other places. The reuse of the ClassNav component shouldn't be too hard and I can do it later when we have a better idea of what the items to be passed into renderStudentSort from the popup view should be. Also I don't have the redux action/name for the items yet.

PR #186, opened 2020-06-18
Question icons

mklewandowski:
This PR adds question icons and a new method for importing and using SVG icons.
- Add svgr/webpack for importing SVG files
- Add new svg icons, remove icon from index.html
- Add question utilities and use new icons in level viewer

At present the following image question types are handled:
-Image Question with Open Response (default for image questions)
-Interactive
-Multiple Choice Scored
-Multiple Choice NonScored
-Open Response

And as we rough things in further, we can add handling for these:
-Fill in the Blank
-Image Question
-Interactive Codap
-Interactive Table
-Multiple Answer Scored
-Multiple Answer NonScored


js/global.d.ts
```diff
@@ -1,2 +1,6 @@
 declare module "*.less";
 declare module "*.png";
+declare module "*.svg" {
+  const content: any;
+  export default content;
```
kswenson, line 5: Out of curiosity, what's the significance of these `content` declarations?

mklewandowski: I modeled it after how it was handled in GeoCode.  Let me see if it behaves the same without them.

mklewandowski: Works fine without it.  Maybe originated from somewhere like here?
https://stackoverflow.com/questions/54121536/typescript-module-svg-has-no-exported-member-reactcomponent
I'm going to remove it since it seems unnecessary.

PR #191, opened 2020-06-19
Handle some cases when question/answer data is corrupted

pjanik:
https://www.pivotaltracker.com/story/show/173400719

I ran the same activity that Kevin reported to be broken. But I didn't have any problems (I ran it using multiple student accounts). Then, I asked Kevin to send me his broken report link. I started debugging and found issues that are fixed by this PR change.

I think that Kevin had to run this activity as a student before we added metadata support. So, some connections between answers and questions were wrong. E.g. "iframe_interactive" response type was referencing "multiple_choice" question type. That's why the code couldn't find there "selectedChoices" property.

Other issue was about MC questions themselves. Some were not published using Authored Metadata approach, but answers to them were already using Runtime Metadata. Changes in report-tree fixed that / prevented the crash. Although I could later fix it just by editing these questions in authoring mode (so they got republished), even without these code changes (I tested that locally).


pjanik:
@scytacki, I've added two things to test data:
- MC question without choices
- MC answer pointing to deleted / nonexistent choice
(as they're causing problems before in Kevin's report)

I've also updated the code a bit to be more explicit and show to teacher that MC question doesn't have any choices or student has selected a choice that later has been deleted. I used these labels in Cypress tests. I also realized that GEODE Dashboard and Portal Dashboard were affected, so some of the new changes are related to that.

scytacki:
Looks good to me.
I'm going to have @eireland review it since I think she updated the mock data recently.

PR #192, opened 2020-06-22
Update icons

mklewandowski:
This PR removes the remaining SVG icons from index.html and moves them to standalone SVG files.  A handful of CSS classes are also updated to add pointer icons or icon classes which were removed in the previous PRs.


js/components/portal-dashboard/account-owner.tsx
```diff
@@ -12,7 +12,7 @@ export class AccountOwnerDiv extends React.PureComponent <IProps> {
       <div className={css.accountOwner} data-cy="account-owner">
         <div className={css.accountOwnerName}>{this.props.userName}</div>
         <div className={css.accountOwnerIcon}>
```
eireland, line 14: Remove class name for div since there is no accountOwnerIcon in css

mklewandowski: Good catch.  That whole div can come out, in fact.  Fixed.

mklewandowski:
@eireland Thanks!  I made the fix that you requested.  Good catch!  If you approve, we can merge this in once it builds.

PR #195, opened 2020-06-23
Add sequence name to header

eireland:


css/portal-dashboard/custom-select.less
```diff
@@ -7,8 +7,6 @@
     display: flex;
     flex-direction: row;
     align-items: center;
-    width: 200px;
-    height: 30px;
```
mklewandowski, line 11: From what I can tell we still need this `height: 30px;` to match the spec.  Otherwise the header of the select shrinks vertically. Studying the spec, I think we also need a `box-sizing: border-box;` to get it to be pixel perfect.  So I think this class needs:
```
height: 30px;
box-sizing: border-box;
```

css/portal-dashboard/class-nav.less
```diff
@@ -10,9 +10,13 @@
 
   .chooseClass {
     margin: 20px 0 15px 19px;
+    width: 200px;
+    height: 30px;
```
mklewandowski, line 14: I think you can get away with not adding these.  The width here is actually constraining the width of the custom select header so it means the custom select needs a parent container to control the width.  I think it's better to be able to place a custom component anywhere without having to add a styled parent to control it's layout.

css/portal-dashboard/class-nav.less
```diff
@@ -10,9 +10,13 @@
...
   }
 
   .studentSort {
     margin-left: 19px;
+    width: 200px;
+    height: 30px;
```
mklewandowski, line 20: Same as above.

css/portal-dashboard/custom-select.less
```diff
@@ -75,6 +72,14 @@
...
     pointer-events: none;
 
+    &.classNav {
+      width: 200px;
+    }
+
+    &.assignmentList {
+      width: 300px;
+    }
+
```
mklewandowski, line 82: I would make the width: 200px the default value and then just use width: 300px with a `wide` class to keep this component generic and unaware of its context (since it's designed to be a generic, reusable control).  So I would just do:
```
.list
  width: 200px;
 ...

    &.wide{
      width: 300px;
    }
...
```

mklewandowski: Then add the same to the `header` class.

css/portal-dashboard/header.less
```diff
@@ -36,11 +36,26 @@
...
+    height: 100%;
+    font-size: 16px;
+
+    .assignmentTitle{
+      margin-right: 10px;
+      padding-bottom:3px;
+    }
+    .chooseAssignment{
+      width: 300px;
+      height: 30px;
```
mklewandowski, line 56: I'm guessing similarly we won't need this width and height if we restructure the custom select.

js/components/portal-dashboard/class-nav.tsx
```diff
@@ -41,6 +41,7 @@ export class ClassNav extends React.PureComponent<IProps> {
           trackEvent={trackEvent}
           HeaderIcon={ClassIcon}
           dataCy={"choose-class"}
+          isWide={false}
```
mklewandowski, line 44: I would make this an optional param so we don't need to add it unless we need it.

js/components/portal-dashboard/class-nav.tsx
```diff
@@ -59,6 +60,7 @@ export class ClassNav extends React.PureComponent<IProps> {
           trackEvent={trackEvent}
           HeaderIcon={SortIcon}
           dataCy={"sort-students"}
+          isWide={false}
```
mklewandowski, line 63: Same as above.  If this is an optional param, we don't need to add this,

js/components/portal-dashboard/custom-select.tsx
```diff
@@ -11,6 +11,7 @@ interface IProps {
   trackEvent: (category: string, action: string, label: string) => void;
   HeaderIcon: SvgIcon;
   dataCy: string;
+  isWide: boolean;
```
mklewandowski, line 14: Make optional:
`isWide?: boolean;`

js/components/portal-dashboard/custom-select.tsx
```diff
@@ -65,16 +66,17 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
 
   private renderList = () => {
     const { items } = this.props;
+    const showWide = this.props.isWide ? css.assignmentList : css.classNav;
```
mklewandowski, line 69: If we assume 200px is the default, I would just do:
`const showWide = this.props.isWide ? css.wide: "";`

mklewandowski: And then we need the same in the header.

js/containers/portal-dashboard/portal-dashboard-app.tsx
```diff
@@ -91,10 +91,11 @@ class PortalDashboardApp extends React.PureComponent<IProps, IState> {
     // In order to list the activities in the correct order,
     // they must be obtained via the child reference in the sequenceTree …
     const activityTrees: Map<any, any> | false = sequenceTree && sequenceTree.get("children");
+
     return (
       <div className={css.portalDashboardApp}>
-        <Header userName={userName} setCompact={setCompactReport} />
-        {activityTrees &&
+        {sequenceTree&& <Header userName={userName} setCompact={setCompactReport} sequenceName={sequenceTree.get("name")} trackEvent={trackEvent}/> }
```
mklewandowski, line 97: add space after sequenceTree and before &&

js/components/portal-dashboard/header.tsx
```diff
@@ -19,12 +23,34 @@ export class Header extends React.PureComponent <IProps> {
           <div className={css.pin} />
           <div className={css.title}>Dashboard</div>
         </div>
-        <div className={css.assignment}>Assignment:</div>
+        <div className={css.headerCenter}>
+          <div className={css.assignmentTitle}>
+            Assignment:
+        </div>
```
mklewandowski, line 29: I think this needs one more tab to align correctly.

js/components/portal-dashboard/header.tsx
```diff
@@ -19,12 +23,34 @@ export class Header extends React.PureComponent <IProps> {
...
           <HeaderMenuContainer setCompact={this.props.setCompact} />
         </div>
       </div>
     );
   }
+
+  private renderAssignmentSelect = () => {
+    const { sequenceName, trackEvent } = this.props;
+    return (
+      <div className={`${css.chooseAssignment}`}>
```
mklewandowski, line 44: I think this div can come out.

js/containers/portal-dashboard/portal-dashboard-app.tsx
mklewandowski: And probably put on multiple lines for readability.

PR #196, opened 2020-06-23
Add cypress smoke test for student list

eireland:
Added student response area layout
Added selectors to question area elements
Added visibility tests for question area
Adds the question area to the popup dashboard
Splits question nav from question overlay to reuse it in all responses popup
Made isOverlay optional so default mode is when not in overlay


js/components/portal-dashboard/all-responses-popup/popup-class-nav.tsx
```diff
@@ -7,7 +7,7 @@ import { SpotlightDialog } from "./spotlight-dialog";
 import SortIcon from "../../../../img/svg-icons/sort-icon.svg";
 import StudentViewIcon from "../../../../img/svg-icons/student-view-icon.svg";
 import QuestionViewIcon from "../../../../img/svg-icons/question-view-icon.svg";
-import SpotlightIcon from "../../../../img/svg-icons/question-view-icon.svg";
+import SpotlightIcon from "../../../../img/svg-icons/spotlight-icon.svg";
```
mklewandowski, line 10: This was probably my mistake while converting icons!  Thanks for fixing!

js/components/portal-dashboard/all-responses-popup/popup-student-response-list.tsx
```diff
@@ -0,0 +1,41 @@
...
+  render() {
+    const { students, isAnonymous } = this.props;
+    return (
+      <div className={css.responseTable} data-cy="popup-response-table">
+        {students && students.map((student: any, i: number) => {
+          const formattedName = isAnonymous
+            ? student.get("name")
+            : `${student.get("lastName")}, ${student.get("firstName")}`;
+
+          return (
```
mklewandowski, line 20: I was a little thrown off by seeing this return inside of the render return.  Might be clearer if this is refactored into a function.  renderStudentRow or something similar.

mklewandowski: Also Code Climate has a point here.  Maybe we should make a utility function that takes a student name and returns the formatted name since we're now doing this in multiple places.

mklewandowski: Where we added `question-utils` and `answer-utils`, we could add `student-utils` and start with this function:
```
export const getFormattedStudentName = (isAnonymous: boolean, student: Map<any, any>) => {
  return (
    isAnonymous
      ? student.get("name")
      : `${student.get("lastName")}, ${student.get("firstName")}`
  );
};
```

Then we could call the utility function in both `StudentNames` and `PopupStudentResponseList`:
`const formattedName = getFormattedStudentName(isAnonymous, student);`

What do you think?

eireland: Was thinking the same thing but didn't know where to put it. Will give it a try.

mklewandowski: I would just make a new module called `student-utils` and put it next to the other two mentioned above.  Unless you see a better location.

eireland: Done. Added a util/student-utils.ts which has the above function. Replaced it in both the student-names and popup-student-response-list

eireland:
The student checkboxes don't do anything yet, and the css for it has not been implemented yet. The student dropdown doesn't have any other options yet. I suspect it will be about filtering students on whether or not they have been given feedback and/or scored, which is what the current report has.

js/util/student-utils.ts
```diff
@@ -0,0 +1,7 @@
+export const getFormattedStudentName = (isAnonymous: boolean, student: Map<any, any>) => {
+  return (
+    isAnonymous
+      ? student.get("name")
+      : `${student.get("lastName")}, ${student.get("firstName")}`
+  );
+};
```
mklewandowski, line 7: Add newline.

PR #198, opened 2020-06-25
Remove hard-coded SVG fill colors

mklewandowski:
Remove hard-coded SVG fill colors so we can set fill color via CSS class.


mklewandowski:
Good question.  I think Michael's goal was to have us import each SVG from the UI spec and use them as they were configured there.  In this case, however, we were going to need to import 11 new assets that were identical to existing assets in everything other than fill color (this was for some work that Evangeline is doing).  That felt like overkill and adding bloat to the project. 
 However, maybe we should have stripped the fill from the beginning on every SVG asset?  At this point doing so would also require adding a bunch of CSS to existing "finished" components.  I'm hoping this is a middle ground that will help speed up Evangeline's work as she needs access to more icon states.

kswenson:
So it sounds like we're saying that some SVGs should keep their fills on import while others should have them stripped. Perhaps we could set up two rules in webpack along with a naming convention so that files named `*.svg` would preserve their fill while files named `*.nofill.svg` would strip theirs or something like that.

mklewandowski:
We could.  I'm hoping that this was a one-off change, but Evangeline can speak more to if there are many more icons/SVGs that still need to come in.

kswenson:
To be clear, I'm not thinking of something that should hold up this PR. I'm just thinking that if we're going to standardize on this approach for CC projects, we should have a solution that doesn't involve hand-editing a subset of the SVGs if we can avoid it.

mklewandowski:
Ah, right.  It could be a nice addition to have something like this.  Although usually Michael gives us SVGs with the fill stripped off if the fill is going to change.  I think this was a rare case where he didn't have time to give us the final SVGs and we were forced to export them from Zeplin instead.  Also, I wish we could standardize SVG usage in the starter-projects - I think that would be very useful since 95% of our projects need SVG icons.

mklewandowski:
Anyway, I'm going to merge, but I think this should be a part of the conversation about updating the starter-projects.

PR #203, opened 2020-08-25
174390612 hide feedback and class response UI elements

mklewandowski:
This PR removes the feedback toggle in the upper left nav area, the feedback button in the student responses popup, and the class response section in the question overlay as we prepare for an MVP version of the new portal dashboard.  Code has been commented out in several cases with a "Removed for MVP" tag so we can more easily add it back in later.


eireland:
Would it be better to skip the tests for now instead of commenting them out?

mklewandowski:
@eireland I made changes based on our discussion.  Let me know if you see anything else.  Otherwise, I'll merge once approved.

PR #205, opened 2020-09-01
Add dynamic image answer component

mklewandowski:
Add new image answer component that can be used in the question overlay and in the all student responses window.  Component needs to be dynamically sizable or have static dimensions based on its context.


js/components/portal-dashboard/answers/image-answer.tsx
```diff
@@ -0,0 +1,83 @@
...
+    React.useLayoutEffect(() => {
+      setSize(target.current.getBoundingClientRect());
+    }, [target]);
+
+    useResizeObserver(target, (entry: any) => setSize(entry.contentRect));
+    return size;
+  };
+
+  const divTarget = React.useRef(null);
+  const divSize: any = useSize(divTarget);
```
kswenson, line 40: Can't we do better than `any` here?

js/components/portal-dashboard/answers/image-answer.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const containerHeight: number = staticSize ? kStaticHeight: divSize && divSize.height;
+  const containerWidth: number = staticSize ? kStaticWidth : divSize && divSize.width;
```
kswenson, line 42: Could use optional chaining here.

mklewandowski: Fixed.

js/components/portal-dashboard/answers/image-answer.tsx
mklewandowski: Hmmm.... perhaps a `ClientRect`, but then it can also be undefined.  Which if the undefined is added to the type it results in the "property doesn't exist on type never" warning.

mklewandowski: Actually `DOMRectReadOnly` might be the right choice.  But I still have the "type never" warning to get around.

mklewandowski: `const divSize: DOMRect | DOMRectReadOnly = useSize(divTarget);` works (the types that we can get from `target.current.getBoundingClientRect()` and `entry.contentRect`) once I also specified an initial state inside of `useSize`.

PR #209, opened 2020-09-08
174514777 image lightbox

eireland:
Opens answer image modal/lightbox when thumbnail is clicked


css/portal-dashboard/answers/answer-modal.less
```diff
@@ -0,0 +1,132 @@
+@import "../variables";
+
+.modal-dialog .modal-content {
+  border-radius: 8px !important;
+}
```
mklewandowski, line 5: Is this unneeded now?

css/portal-dashboard/answers/answer-modal.less
```diff
@@ -0,0 +1,132 @@
...
+
+.lightbox {
+  width: 640px;
+  border-radius: 8px !important;
```
mklewandowski, line 9: Unneeded?

js/components/portal-dashboard/answers/image-answer.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useState }  from "react";
 // TODO: replace with new model
```
mklewandowski, line 2: We can remove this comment now.

js/components/portal-dashboard/answers/answer-modal.tsx
```diff
@@ -0,0 +1,76 @@
...
+                {this.renderAnswer(type)}
+              </div>
+              <div className={css.responseText} data-cy="response-note">
+                {answerText}
+              </div>
+            </div>
+          </Modal.Body>
+      </Modal>
+    );
+  }
```
mklewandowski, line 59: This could be refactored into render header and body functions.

css/portal-dashboard/answers/answer-modal.less
eireland: Took off the !important. But still need the border-radius.

mklewandowski (approved):
👍 I think this looks good.  I made a couple comments where things can be removed, and I think you could refactor the render function to avoid the code climate warning.  I would also add a comment about why we needed to import the CSS file with the special class rules for the modal.

js/components/portal-dashboard/answers/answer-modal.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { PureComponent } from "react";
+import { Modal } from "react-bootstrap";
+import SmallCloseIcon from "../../../../img/svg-icons/small-close-icon.svg";
+import { QuestionTypes } from "../../../util/question-utils";
+import { renderHTML } from "../../../util/render-html";
+
+import css from "../../../../css/portal-dashboard/answers/answer-modal.less";
+import "../../../../css/portal-dashboard/answers/answer-modal.css";
```
mklewandowski, line 8: I would add a comment about why this css import is needed.

mklewandowski: Something about how the CSS module import mutates the class names so that they are dynamic, but we needed to make a change to a static class name (`modal-content`) that exists in the imported modal dialog library.  The CSS import does that.  It lets us specify style on a statically named CSS class.

PR #210, opened 2020-09-08
174712393 add drawing prompt

eireland:
Adds the drawing prompt to the image question prompt question prompt to question area in both overlay and all student response view

### DIFF
--- a/js/components/portal-dashboard/answers/answer-modal.tsx
+++ b/js/components/portal-dashboard/answers/answer-modal.tsx
@@ -2,7 +2,7 @@ import React, { PureComponent } from "react";
 import { Modal } from "react-bootstrap";
 import SmallCloseIcon from "../../../../img/svg-icons/small-close-icon.svg";
 import { QuestionTypes } from "../../../util/question-utils";
-import { renderHTML } from "../../../util/render-html";
+import striptags from "striptags";
 
 import css from "../../../../css/portal-dashboard/answers/answer-modal.less";
 // needed to specify modal styles that can only be access through the react-bootstrap modal-content class.
@@ -24,7 +24,7 @@ export class AnswerModal extends PureComponent<IProps> {
     const type = answer?.get("questionType");
     const questionType = QuestionTypes.find(qt => qt.type === type);
     const QuestionIcon = questionType?.icon;
-    const prompt = question && renderHTML(question.get("prompt"));
+    const prompt = question && striptags((question.get("drawingPrompt")).replace(/&nbsp;/g,' ')) + striptags((question.get("prompt")).replace(/&nbsp;/g,' '));
 
     if (!answer) { return null; }
     return (
@@ -34,7 +34,7 @@ export class AnswerModal extends PureComponent<IProps> {
           <div className={css.questionPrompt} data-cy="question-prompt">
             <QuestionIcon className={`${css.icon} ${css.questionTypeIcon}`} />
             <span className={css.questionPromptText} data-cy="question-prompt-text">
-              {prompt}
+              {prompt && prompt}
             </span>
           </div>
           <div className={css.contentArea} data-cy="lightbox-content-area">
--- a/js/components/portal-dashboard/answers/answer-modal.tsx
+++ b/js/components/portal-dashboard/answers/answer-modal.tsx
@@ -24,7 +24,7 @@ export class AnswerModal extends PureComponent<IProps> {
     const type = answer?.get("questionType");
     const questionType = QuestionTypes.find(qt => qt.type === type);
     const QuestionIcon = questionType?.icon;
-    const prompt = question && striptags((question.get("drawingPrompt")).replace(/&nbsp;/g,' ')) + striptags((question.get("prompt")).replace(/&nbsp;/g,' '));
+    const prompt = question? striptags((question.get("drawingPrompt")).replace(/&nbsp;/g,' ')) + striptags((question.get("prompt")).replace(/&nbsp;/g,' ')) : "";
 
     if (!answer) { return null; }
     return (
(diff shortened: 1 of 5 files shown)


js/components/portal-dashboard/question-area.tsx
```diff
@@ -48,8 +49,10 @@ export class QuestionArea extends React.PureComponent<IProps>{
         </div>
         <div className={`${css.questionTextArea} ${!useMinHeight ? css.minHeight : ""}`} data-cy="question-content">
           <div className={css.questionText}>
-            {prompt ? striptags(prompt.replace(/&nbsp;/g,' ')) : ""}
+          {prompt ? renderHTML(prompt) : ""}
+            {/* {prompt ? striptags(prompt.replace(/&nbsp;/g,' ')) : ""} */}
```
mklewandowski, line 53: Is this commented piece obsolete?  Probably want to remove it.  Also want to tab in the `{prompt ? renderHTML(prompt) : ""}` line so it is properly aligned.

js/components/portal-dashboard/question-area.tsx
```diff
@@ -48,8 +49,10 @@ export class QuestionArea extends React.PureComponent<IProps>{
...
           </div>
+          { drawingPrompt && this.renderImageQuestionPrompt(drawingPrompt)}
```
mklewandowski, line 55: Tab in for proper alignment.

js/components/portal-dashboard/question-area.tsx
```diff
@@ -77,4 +80,12 @@ export class QuestionArea extends React.PureComponent<IProps>{
       </div>
     );
   }
+
+  private renderImageQuestionPrompt = (drawingPrompt: string) => {
+    return (
+      <div className={css.imageQuestionPrompt}>
+        {renderHTML(drawingPrompt)}
+      </div>
```
mklewandowski, line 88: One minor issue is that if there is only a drawing prompt and no main prompt, then they will still see the divider line.  Not sure if that matters.

js/components/portal-dashboard/question-area.tsx
mklewandowski: And you can rewrite simply as `{ prompt && renderHTML(prompt) }`

mklewandowski (approved):
👍 LGTM!  I added a couple of alignment/code style suggestions.  The only thing you might consider is ONLY showing the divider line if there is BOTH a prompt and a drawing prompt (which I would do as a 1px height div in that case instead of a top border or if you don't want to go that route, you could add a CSS class to hide the border).  If there is just a drawing prompt, we show the divider (since it is displayed as a border on the drawing prompt container).

mklewandowski (changes requested):
Ah wait, shouldn't the drawing prompt be the one that is on top?  I just reread the story specs, and I think the drawing prompt goes first and the main prompt second.

js/components/portal-dashboard/question-area.tsx
mklewandowski: Looking  more closely at the story spec, the drawing prompt goes first.  This should be `{ drawingPrompt && renderHTML(drawingPrompt ) }`

js/components/portal-dashboard/question-area.tsx
mklewandowski: Looking more closely at the story spec, the prompt goes second.  This should be `{ prompt && this.renderImageQuestionPrompt(prompt) }`

js/components/portal-dashboard/question-area.tsx
```diff
@@ -77,4 +80,12 @@ export class QuestionArea extends React.PureComponent<IProps>{
       </div>
     );
   }
+
+  private renderImageQuestionPrompt = (drawingPrompt: string) => {
```
mklewandowski, line 84: argument should be `prompt`, not `drawingPrompt`

js/components/portal-dashboard/question-area.tsx
```diff
@@ -24,6 +24,7 @@ export class QuestionArea extends React.PureComponent<IProps>{
     const questionType = QuestionTypes.find(qt => qt.type === type && qt.scored === scored);
     const QuestionIcon = questionType?.icon;
     const mcChoices: Map<any,any> = currentQuestion?.get("choices") || [];
+    const drawingPrompt = currentQuestion?.get("drawingPrompt") || undefined;
```
mklewandowski, line 27: Can be just `const drawingPrompt = currentQuestion?.get("drawingPrompt");`

js/components/portal-dashboard/question-area.tsx
mklewandowski: And you could pass a boolean here like 'showDivider` so you only show the divider if both prompts exist.

js/components/portal-dashboard/question-area.tsx
mklewandowski: If you had a `showDivider` boolean, then you could do:
```
<div className={`${css.imageQuestionPrompt} ${showDivider ? css.showDivider : ""}`}>
```
And add the appropriate CSS to show/hide the border

js/components/portal-dashboard/answers/answer-modal.tsx
```diff
@@ -34,7 +34,7 @@ export class AnswerModal extends PureComponent<IProps> {
           <div className={css.questionPrompt} data-cy="question-prompt">
             <QuestionIcon className={`${css.icon} ${css.questionTypeIcon}`} />
             <span className={css.questionPromptText} data-cy="question-prompt-text">
-              {prompt}
+              {prompt && prompt}
```
mklewandowski, line 37: This is odd.  Isn't prompt a string?  Shouldn't this just be `prompt` and then we determine above if it's empty string?

mklewandowski: I think we can safely just put in `{prompt}`.  Looks like in the worst case we just have an empty string "" that we will put in the span.

mklewandowski: I think we can safely just put in `{prompt}`.  In the worst case prompt will just have an empty string, I believe.

mklewandowski (approved):
👍 LGTM!  I would just clean up the awkward `{prompt && prompt}` and just use `{prompt`} while confirming that prompt is always a string.  Based on the limited space, I think you're right that it's fine to use striptags in the modal for the prompts while showing any tags in the large space that we have in the question overlay.

PR #211, opened 2020-09-09
Set min-size and add vertical scrolling to answer component

mklewandowski:
Add a minimum size to the answer component and add scroll bars when needed so content is always displayed in question overlay panel.
resize


eireland (approved):
Looks good

PR #212, opened 2020-09-10
174390484 interactive iframe

mklewandowski:
This PR makes fixes to the handling of the interactive iframe answers and some small UI improvements to the all student responses overlay.  Changes include: 
- Correctly determine if interactive can be opened with "View Work" link or only with "View work in new tab" link.
- Better handling of interactive sizing in question overlay (improvements to both the image answer and the interactive answer).
- Fix the min height of answers in the all student responses overlay (min height was being applied at all times which was adding lots of unneeded whitespace).
- Fix background color in upper left nav in all student responses (fixes the white band that sometimes appeared at bottom of nav component).
- Fix student name column width in all student responses (the column width did not line up with the upper left nav component).

### DIFF
--- a/css/portal-dashboard/all-responses-popup/popup-class-nav.less
+++ b/css/portal-dashboard/all-responses-popup/popup-class-nav.less
@@ -8,6 +8,7 @@
 .popupClassNav {
   width: 240px;
   font-weight: normal;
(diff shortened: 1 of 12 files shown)


eireland (approved):
LGTM!

PR #213, opened 2020-09-10
Use ellipsis when level viewer activity text is expanded and overflows

mklewandowski:
Use ellipsis when the level viewer activity text is expanded and overflows. There are probably other places in the app where this is needed, but this gives us a starting point.
image


eireland (approved):
LGTM!

PR #214, opened 2020-09-10
Fix layout issues in student name container

mklewandowski:
This small PR fixes the following student name display issues (might need to switch classes to test):
- long names are not displayed properly in response table
- long names cause selection bubble to format incorrectly in all student response overlay

image


eireland (approved):
LGTM!

PR #215, opened 2020-09-14
Remove menu options, add help URL

mklewandowski:
Updates to the upper right menu for the portal dash MVP.  Changes include:
- remove print and download from menu
- add help link that opens in new tab when help is clicked


eireland (approved):
LGTM!

PR #216, opened 2020-09-14
Response table selection

mklewandowski:
This PR fixes a bug/unimplemented feature in the response table.  When an response table icon is clicked, the current question and student in the question overlay need to change state to reflect the item that was clicked on.  This PR adds a current student index to the to redux state so that we can access a current student in parallel components.  Otherwise, existing redux state is used to manage clicks in the response table.

tableclick

### DIFF
--- a/js/components/portal-dashboard/student-answers.tsx
+++ b/js/components/portal-dashboard/student-answers.tsx
@@ -6,6 +6,8 @@ import css from "../../../css/portal-dashboard/student-answers.less";
 interface IProps {
   activities: any;
   currentActivity: any;
+  currentQuestion: any;
+  currentStudentIndex: number;
   expandedActivities: any;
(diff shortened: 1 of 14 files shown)


js/components/portal-dashboard/student-answers.tsx
```diff
@@ -113,4 +141,5 @@ export class StudentAnswers extends React.PureComponent<IProps> {
       <div className={`${css.progressIcon} ${cssClass}`}/>
     );
   }
+
```
eireland, line 144: extra space? line 144

mklewandowski: That's just the spacing format that I prefer.  I tend to put an empty line before/after/between functions for readability.

eireland (approved):
LGTM!

PR #217, opened 2020-09-15
Add interactive mock data

eireland:


mklewandowski (approved):
👍 Looks good to me.  If you merge this, I can rebase my branch that fixes the load of the interactive answers in the question detail and then we can see if it is truly working.

PR #218, opened 2020-09-15
Fix interactive load error when navigating from one interactive to another

mklewandowski:
This PR adds unique keys to the answer containers in the question overlay.  This should guarantee that when navigating to a new student or question, React will flush the answer node and creates a new component where we can properly initialize the interactive iframe.  This is hard to verify without the new mock data.  However, barring any obvious issues, I suggest that we merge this and then rebase the mock data branch onto master where we can confirm the fix.


eireland (approved):
Merge it!

PR #219, opened 2020-09-15
174800482 close popup

eireland:
This add the ability to close the all response pop up when clicking on the Activity name on the top left.


mklewandowski (approved):
👍 LGTM!

PR #220, opened 2020-09-15
Improvements to the activity score

mklewandowski:
This PR adds minor improvements to the score that appears in the level viewer:
- prevent score from behaving as button.  Remove hover, active, and cursor states
- better positioning of score check

image


eireland (approved):
merge it!

PR #221, opened 2020-09-15
Sync anonymize toggles

mklewandowski:
This PR syncs the anonymize toggle in the upper left nav and in the all student responses view so that they always display the same on/off state.


eireland (approved):
Merge it!

PR #222, opened 2020-09-16
174654525 fill blank question detail

eireland:
Adds ability to show question prompts for scaffolded questions and fill in the blank questions. 
Added mock data to reflect that prompts are part of the data sent to the report as part of the question map.
Refactored how questions are rendered in the question area so each type is now a separate component like the answers.


css/portal-dashboard/questions/multiple-choice-question.less
```diff
@@ -0,0 +1,27 @@
...
+    margin-right: 8px;
+  }
+  .correct {
+    font-weight: bold;
+  }
+}
+
+.questionText {
+  margin: 0px 20px 10px 20px;
+}
```
mklewandowski, line 27: newline

css/portal-dashboard/question-area.less
```diff
@@ -20,9 +20,10 @@
     display: flex;
     flex-direction: row;
     align-items: center;
-    flex: 1;
+    flex-grow: 2;
```
mklewandowski, line 23: Why is this needed?  `flex: 1` sets `flex-grow` to 1.  This change will make this box grow at 2 times the rate of other flex boxes in the same container.  I'm wondering if the better approach is to leave this as 1, and then make it so the `rightIcons` class doesn't grow, by using `flex: auto` or setting a width and using `flex-grow: 0`.  I think you might get better results with that.

js/components/portal-dashboard/questions/multiple-choice-question.tsx
```diff
@@ -0,0 +1,44 @@
...
+export default class MultipleChoiceQuestion extends PureComponent<IProps> {
+  render() {
+    const { question } = this.props;
+    const prompt = question?.get("prompt");
+    const choices: Map<string,any> = question?.get("choices") || [];
+
+    return (
+      <div className={css.questionText}>
+      {prompt && renderHTML(prompt)}
+      { choices.size > 0 ? choices.toArray().map(this.renderMultipleChoiceChoices(choices.toArray().length)) : "" }
```
mklewandowski, line 20: Tab in the content inside the div for consistency and readability.

js/containers/portal-dashboard/answer.tsx
```diff
@@ -22,22 +22,21 @@ class Answer extends React.PureComponent<AnswerProps> {
     const scored = question && question.get("scored");
     const questionType = QuestionTypes.find(qt => qt.type === qtype && qt.scored === scored);
     const answerType = getAnswerType(answer, question);
```
mklewandowski, line 24: Removed unused `answerType` (just whack the entire line).

js/components/portal-dashboard/questions/multiple-choice-question.tsx
```diff
@@ -0,0 +1,44 @@
...
+  private renderMultipleChoiceChoices = (numChoices: number) => (choices: Map<string, any>, i: number) => {
+    let multipleChoiceContent, multipleChoiceContentClass;
+    if (choices.get("correct")) {
+      multipleChoiceContentClass = `${css.mcContent} ${css.correct}`;
+      multipleChoiceContent = choices.get("content") + " (correct)";
+    }
+    else {
+      multipleChoiceContentClass = `${css.mcContent}`;
+      multipleChoiceContent = choices.get("content");
+    }
```
mklewandowski, line 34: Maybe replace the `let`  and the `if else` with:
```
    const multipleChoiceContent = `${choices.get("content")} ${choices.get("correct") ? "(correct)" : ""}`;
    const multipleChoiceContentClass = `${css.mcContent} ${choices.get("correct") ? css.correct : ""}`;
```

js/components/portal-dashboard/questions/multiple-choice-question.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    return (
+      <div className={css.questionText}>
+      {prompt && renderHTML(prompt)}
+      { choices.size > 0 ? choices.toArray().map(this.renderMultipleChoiceChoices(choices.toArray().length)) : "" }
+      </div>
+    );
+  }
+
+  private renderMultipleChoiceChoices = (numChoices: number) => (choices: Map<string, any>, i: number) => {
```
mklewandowski, line 25: I don't think you need this function currying.  Can just be:
`private renderMultipleChoiceChoices = (choices: Map<string, any>, i: number) => {`

js/components/portal-dashboard/questions/multiple-choice-question.tsx
mklewandowski: Can be rewritten as: 
`{ choices.toArray?.().map(this.renderMultipleChoiceChoices) }`

js/components/portal-dashboard/questions/multiple-choice-question.tsx
```diff
@@ -0,0 +1,44 @@
...
+      multipleChoiceContentClass = `${css.mcContent} ${css.correct}`;
+      multipleChoiceContent = choices.get("content") + " (correct)";
+    }
+    else {
+      multipleChoiceContentClass = `${css.mcContent}`;
+      multipleChoiceContent = choices.get("content");
+    }
+    return (
+      <div className={css.choiceWrapper} key={`choices ${i}`}>
+        <div className={`${css.choiceIcon}`} />
```
mklewandowski, line 37: `<div className={css.choiceIcon} />`

js/components/portal-dashboard/questions/multiple-choice-question.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <div className={`${multipleChoiceContentClass}`}>
```
mklewandowski, line 38: `<div className={multipleChoiceContentClass}>`

js/components/portal-dashboard/questions/open-response-question.tsx
```diff
@@ -0,0 +1,22 @@
...
+    const { question } = this.props;
+    const prompt = question?.get("prompt");
+
+    return (
+      <div className={css.questionText}>
+        {prompt && renderHTML(prompt)}
+      </div>
+    );
+  }
+}
```
mklewandowski, line 22: I'll just note that these question components would be good choices to turn into functional components.

js/containers/portal-dashboard/question.tsx
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import { Map } from "immutable";
+import MultipleChoiceQuestion from "../../components/portal-dashboard/questions/multiple-choice-question";
+import OpenResponseQuestion from "../../components/portal-dashboard/questions/open-response-question";
+import  ImageQuestion from "../../components/portal-dashboard/questions/image-question";
```
mklewandowski, line 5: Extra space sneaked in here after `import`

js/containers/portal-dashboard/question.tsx
```diff
@@ -0,0 +1,46 @@
...
+    return (
+      <div className={css.questionText} data-cy="current-question">
+          {this.renderQuestion(qtype)}
+      </div>
+    );
+  }
+
+  renderQuestion = (type: string) => {
+    const { question } = this.props;
+    const QuesitonComponent: any = {
```
mklewandowski, line 28: `QuestionComponent`

js/containers/portal-dashboard/question.tsx
```diff
@@ -0,0 +1,46 @@
...
+      "multiple_choice": MultipleChoiceQuestion,
+      "open_response": OpenResponseQuestion,
+      "image_question": ImageQuestion,
+      "iframe_interactive": IframeQuestion,
+    };
+    const QComponent = (question && QuesitonComponent[type]);
```
mklewandowski, line 34: `QuestionComponent`

js/containers/portal-dashboard/question.tsx
```diff
@@ -0,0 +1,46 @@
...
+      return (
+        <div>Question type not supported.</div>
+      );
+    }
+    else {
+      return (
+        <QComponent question={question} />
+      );
+    }
+  }
```
mklewandowski, line 45: I would remove the if/else and two returns and use:
```
    return (
      <React.Fragment>
      { QComponent 
        ? <QComponent question={question} />
        : <div>Question type not supported.</div>
      }
      </React.Fragment>
    );
```

mklewandowski (changes requested):
Looks good!  Almost there.  I made comments for some code clean-up.  I would also investigate the `flex-grow: 2` and try to keep that as `flex: 1` while instead changing the `rightIcons` class to be a static width (use `flex-grow: 0` and set width).  It should prevent some unexpected results and perform better.  Let me know if that makes sense.  The other changes should be in my comments.

js/containers/portal-dashboard/question.tsx
kswenson: Since there's only a single child you shouldn't need the `<React.Fragment>` either.

mklewandowski: Yes. Can be:
```
    return (
       QComponent
        ? <QComponent question={question} />
        : <div>Question type not supported.</div>
    );
```

css/portal-dashboard/question-area.less
eireland: Set the width for the left side title text, and positioned the right side icons.

js/components/portal-dashboard/level-viewer.tsx
```diff
@@ -8,7 +8,7 @@ import LaunchIcon from "../../../img/svg-icons/launch-icon.svg";
 import css from "../../../css/portal-dashboard/level-viewer.less";
 
 // from level-viewer.less
-const progressWidth = 120;
+const progressWidth = 116;
```
mklewandowski, line 11: Per my comments in Slack, I'm voting for this to be 117.

css/portal-dashboard/questions/multiple-choice-question.less
mklewandowski: Might still want to add a newline at the end of this file to be consistent.  I know we haven't enforced this as much in CSS modules, but seems like a good practice.

js/containers/portal-dashboard/answer.tsx
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import { getAnswersByQuestion } from "../../selectors/report-tree";
 import { connect } from "react-redux";
-import { getAnswerType, getAnswerIconId, AnswerProps } from "../../util/answer-utils";
+import { getAnswerType, AnswerProps } from "../../util/answer-utils";
```
mklewandowski, line 4: `getAnswerType` is no longer used and throws a lint warning.  We can remove it from the imports.

js/components/portal-dashboard/questions/multiple-choice-question.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  //   return (
+  //     <div className={css.choiceWrapper} key={`choices ${i}`}>
+  //       <div className={css.choiceIcon} />
+  //       <div className={multipleChoiceContentClass}>
+  //         {multipleChoiceContent}
+  //       </div>
+  //     </div>
+  //   );
+  // }
```
mklewandowski, line 49: Remove obsolete code.

js/components/portal-dashboard/questions/multiple-choice-question.tsx
```diff
@@ -0,0 +1,50 @@
...
+        </div>
+      );
+    };
+
+    return (
+      <div className={css.questionText}>
+        {prompt && renderHTML(prompt)}
+        { choices.toArray?.().map(renderMultipleChoiceChoices) }
+      </div>
+    );
```
mklewandowski, line 34: Might be clearer to try to collapse this so all the render is in one return.  I know that's something that @kswenson mentioned a couple times about many of our React class components - that we have sprawling render functions that are hard to follow.

js/components/portal-dashboard/question-area.tsx
```diff
@@ -17,23 +17,20 @@ export class QuestionArea extends React.PureComponent<IProps>{
...
-    const showDivider = drawingPrompt && prompt;
 
     return (
       <div className={`${css.questionArea} ${hideQuestion ? css.hidden : ""}`}>
         <div className={css.questionTypeHeader}>
           <div className={css.leftTitle}>
             <QuestionIcon className={`${css.icon} ${css.questionTypeIcon}`} />
             <span className={css.questionTypeTitle} data-cy="question-title">
-              {typeText}
+              {type === "iframe_interactive"? (interactiveName? interactiveName : typeText) : typeText}
```
mklewandowski, line 33: I think this will be more clear if in line 23 you have:
`const interactiveName = type === "iframe_interactive" && currentQuestion?.get("name");`

And then here in the render you just use::
`{ interactiveName || typeText }`

This way, you're more fully defining `interactiveName` based on both the type and the question and the render portion becomes more clear and doesn't require two ternaries.

js/components/portal-dashboard/question-area.tsx
```diff
@@ -17,23 +17,20 @@ export class QuestionArea extends React.PureComponent<IProps>{
     const { currentQuestion, hideQuestion, useMinHeight } = this.props;
     const teacherEditionButtonClasses = css.teacherEditionButton;
     const teacherEditionBadge = css.teacherEditionBadge;
-    const prompt = currentQuestion?.get("prompt");
     const type = currentQuestion?.get("type");
     const scored = currentQuestion?.get("scored");
     const typeText = type && type.replace(/_/gm, ' ');
+    const interactiveName = currentQuestion?.get("name");
```
mklewandowski, line 23: Per my comment below:
`const interactiveName = type === "iframe_interactive" && currentQuestion?.get("name");`

mklewandowski (changes requested):
Almost there!  Just a few final tweaks:
- add newline at the end of .less module
- eliminate double ternary in render in `question-area.tsx` (just change two lines per my comments)
- remove obsolete code in `multiple-choice-question.tsx` 
- refactor render in `multiple-choice-question.tsx` per my suggestion
- remove unused import in `answer.tsx` that throws lint warning

js/components/portal-dashboard/questions/multiple-choice-question.tsx
mklewandowski: Here, this should do the trick:
```
    return (
      <div className={css.questionText}>
        { prompt && renderHTML(prompt) }
        { choices.toArray?.().map((choices: Map<string, any>, i: number) => {
            const multipleChoiceContent = `${choices.get("content")} ${choices.get("correct") ? "(correct)" : ""}`;
            const multipleChoiceContentClass = `${css.mcContent} ${choices.get("correct") ? css.correct : ""}`;
            return (
              <div className={css.choiceWrapper} key={`choices ${i}`}>
                <div className={css.choiceIcon} />
                <div className={multipleChoiceContentClass}>
                  {multipleChoiceContent}
                </div>
              </div>
            );
          })
        }
      </div>
    );
```

mklewandowski: Replace your `return` in the functional component with the above.  Then you can remove the `renderMultipleChoiceChoices`.  It will be easier to read and might make code climate happy!

js/components/portal-dashboard/questions/iframe-question.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { PureComponent } from "react";
```
mklewandowski, line 1: Remove unused `PureComponent` import.

js/components/portal-dashboard/questions/image-question.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { PureComponent } from "react";
```
mklewandowski, line 1: Remove unused `PureComponent` import.

js/components/portal-dashboard/questions/open-response-question.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { PureComponent } from "react";
```
mklewandowski, line 1: Remove unused `PureComponent` import.

mklewandowski (approved):
👍 Looks good!  Let's wait for it to build, and then bring it in!

PR #224, opened 2020-09-16
Fix student score calculation

mklewandowski:
Fix the score displayed in the student response table.  Use multiple choice correct/incorrect to determine the score.
image


eireland (approved):
Looks good. Just confirming that we are no longer showing green background if they got all correct?

mklewandowski:
> Looks good. Just confirming that we are no longer showing green background if they got all correct?

I asked about the background highlight in the PT story, and Kiley said this branch looked ok.

PR #225, opened 2020-09-17
Fix handling of scroll state when scrollLeft is 0

mklewandowski:
This PR fixes a bug in the handling of the scroll state.  When the horizontal scroll offset was 0, we were not saving the state due to a faulty comparison.   Instead explicitly test against undefined so that we save the scroll offset when it is 0.

### DIFF
--- a/js/containers/portal-dashboard/portal-dashboard-app.tsx
+++ b/js/containers/portal-dashboard/portal-dashboard-app.tsx
@@ -204,7 +204,7 @@ class PortalDashboardApp extends React.PureComponent<IProps, IState> {
 
   private handleScroll = (e: React.UIEvent<HTMLElement>) => {
     const element = e.target as HTMLElement;
-    if (element && element.scrollLeft !== undefined) {
+    if (element && element.scrollLeft != null) {
       this.setState({ scrollLeft: -1 * element.scrollLeft });
     }
   }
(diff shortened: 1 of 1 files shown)


js/containers/portal-dashboard/portal-dashboard-app.tsx
```diff
@@ -204,7 +204,7 @@ class PortalDashboardApp extends React.PureComponent<IProps, IState> {
 
   private handleScroll = (e: React.UIEvent<HTMLElement>) => {
     const element = e.target as HTMLElement;
-    if (element && element.scrollLeft) {
+    if (element && element.scrollLeft !== undefined) {
```
kswenson, line 207:
```suggestion
    if (element && element.scrollLeft != null) {
```
It's a matter of style/opinion, but I prefer the simpler (to me) test which captures both null and undefined. Otherwise, I'm left wondering what would happen if the value were ever null.

mklewandowski: Yes, that's better.  I was bit by this test already so I don't it to come back and bite me again.

mklewandowski: Fixed.

eireland (approved):
merge it!

PR #226, opened 2020-09-17
Add perfect score green container

mklewandowski:
When a user has all of the scored questions correct, highlight their score with a green box (this was previously part of the UI and removed when we reworked the score).

image


eireland (approved):
merge it!

PR #227, opened 2020-09-17
174831493 horizontal scroll improvements

mklewandowski:
This PR makes changes to the display of the horizontal scroll bar for the response table when the app is at a short width.  The primary change being ensuring that the scrollbar is always visible when the content overflows (previously could be hidden if the vertical scrollbar was also present and the user was scrolled to the top of the container).  Changes include:
- Always show the horizontal scrollbar when it is needed
- Improve responsiveness of app header by adjusting the flex layout - this allows the header to behave better when the app width is small
- Improve the display of the vertical scrollbar and top border of first student row (move cover so we do not occlude these controls)

### DIFF
--- a/css/portal-dashboard/header.less
+++ b/css/portal-dashboard/header.less
@@ -7,7 +7,7 @@
   box-sizing: border-box;
   height: @header-height;
   margin-bottom: 3px;
-  padding: 0 20px 0 20px;
+  padding: 0 20px;
   background-color: @cc-teal-light4;
 
   .appInfo {
--- a/css/portal-dashboard/header.less
+++ b/css/portal-dashboard/header.less
@@ -7,6 +7,7 @@
   box-sizing: border-box;
   height: @header-height;
   margin-bottom: 3px;
+  padding: 0 20px 0 20px;
   background-color: @cc-teal-light4;
 
   .appInfo {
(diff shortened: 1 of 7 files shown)


js/containers/portal-dashboard/portal-dashboard-app.tsx
```diff
@@ -204,7 +197,7 @@ class PortalDashboardApp extends React.PureComponent<IProps, IState> {
 
   private handleScroll = (e: React.UIEvent<HTMLElement>) => {
     const element = e.target as HTMLElement;
-    if (element && element.scrollLeft) {
+    if (element && element.scrollLeft !== null) {
```
kswenson, line 200:
```suggestion
    if (element && element.scrollLeft != null) {
```
Presumably, you want to catch undefined as well.

css/portal-dashboard/header.less
```diff
@@ -7,6 +7,7 @@
   box-sizing: border-box;
   height: @header-height;
   margin-bottom: 3px;
+  padding: 0 20px 0 20px;
```
kswenson, line 10:
```suggestion
  padding: 0 20px;
```
DRY

js/containers/portal-dashboard/portal-dashboard-app.tsx
mklewandowski: Fixed

css/portal-dashboard/header.less
mklewandowski: Fixed

eireland (approved):
merge it!

PR #228, opened 2020-09-17
Fix arrow button states

mklewandowski:
This PR fixes the UI of the show//hide question arrow button in the question overlay so that it matches the UI spec.  Changes include:
- proper arrow orientation.  Arrow points up when question is shown, points down when question is hidden
- arrow has correct color when question is hidden
- arrow has hover states defined in UI spec

See spec for more details:
https://app.zeplin.io/project/5e96f4b8527afa7d28bb1468/screen/5ed4f976e9fe2773c0bd3a33

image

### DIFF
--- a/css/portal-dashboard/question-navigator.less
+++ b/css/portal-dashboard/question-navigator.less
@@ -64,21 +64,36 @@
     color: @cc-charcoal;
   }
 
-  .arrow {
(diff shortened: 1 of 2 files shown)


eireland (approved):
,erge it!

PR #229, opened 2020-09-17
Add activity button colors

mklewandowski:
This PR adds the five unique default activity colors to the activity buttons in the level viewer.

image

### DIFF
--- a/css/portal-dashboard/level-viewer.less
+++ b/css/portal-dashboard/level-viewer.less
@@ -84,7 +84,6 @@
       }
 
       .activityImage {
-        background-color: @cc-teal-light2;
         &.firstColor {
           background-color: @activity-blue;
         }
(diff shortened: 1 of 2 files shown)


css/portal-dashboard/level-viewer.less
```diff
@@ -85,6 +85,21 @@
 
       .activityImage {
         background-color: @cc-teal-light2;
```
eireland, line 87: Why keep this?

eireland (approved):
Except for the one question, it looks good.

css/portal-dashboard/level-viewer.less
mklewandowski: I just wanted a safe fallback color, but I'm ok with taking it out.

PR #230, opened 2020-09-17
174830488 overlay activity link

eireland:
Adds ability to open activity link from the question overlay and all student response popup question area.
Adds ability to open the activity in teacher edition mode if available. If teacher edition mode is not available, teacher edition icon is not shown.

### DIFF
--- a/js/components/portal-dashboard/question-area.tsx
+++ b/js/components/portal-dashboard/question-area.tsx
@@ -11,11 +11,12 @@ interface IProps {
   currentQuestion?: Map<string, any>;
   hideQuestion: boolean;
   useMinHeight?: boolean;
+  hasTeacherEdition: boolean;
 }
 
 export class QuestionArea extends React.PureComponent<IProps>{
   render() {
-    const { currentActivity, currentQuestion, hideQuestion, useMinHeight } = this.props;
+    const { currentActivity, currentQuestion, hideQuestion, useMinHeight, hasTeacherEdition } = this.props;
     const teacherEditionButtonClasses = css.teacherEditionButton;
     const teacherEditionBadge = css.teacherEditionBadge;
     const type = currentQuestion?.get("type");
@@ -39,12 +40,13 @@ export class QuestionArea extends React.PureComponent<IProps>{
             <a className={css.externalLinkButton} href={activityURL} target="_blank" data-cy="open-activity-button">
               <LaunchIcon className={css.icon} />
             </a>
-            <div className={css.teacherEditionIcon}>
-              <a className={teacherEditionButtonClasses} href={activityURL+"?mode=teacher-edition"} target="_blank" data-cy="open-teacher-edition-button">
-                <LaunchIcon className={css.icon} />
-              </a>
-              <div className={teacherEditionBadge}>TE</div>
-            </div>
+            {hasTeacherEdition &&
+              <div className={css.teacherEditionIcon}>
+                <a className={teacherEditionButtonClasses} href={activityURL+"?mode=teacher-edition"} target="_blank" data-cy="open-teacher-edition-button">
+                  <LaunchIcon className={css.icon} />
+                </a>
+               <div className={teacherEditionBadge}>TE</div>
+              </div>}
           </div>
         </div>
         <div className={`${css.questionTextArea} ${!useMinHeight ? css.minHeight : ""}`} data-cy="question-content">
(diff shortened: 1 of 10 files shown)


js/components/portal-dashboard/question-area.tsx
```diff
@@ -34,15 +37,16 @@ export class QuestionArea extends React.PureComponent<IProps>{
...
-                <LaunchIcon className={css.icon} />
-              </a>
-              <div className={teacherEditionBadge}>TE</div>
-            </div>
+            {hasTeacherEdition &&
+              <div className={css.teacherEditionIcon}>
+                <a className={teacherEditionButtonClasses} href={activityURL+"?mode=teacher-edition"} target="_blank" data-cy="open-teacher-edition-button">
+                  <LaunchIcon className={css.icon} />
+                </a>
+               <div className={teacherEditionBadge}>TE</div>
```
mklewandowski, line 48: Very minor, but check spacing.  I think the tabbing is off here and this should be one space in.

mklewandowski: Also, note the this div with the Teacher Edition badge is outside the clickable <a> link.  So even though the div has hover and active effects, clicking on it does nothing.  This threw me off - at first I thought the link didn't work.  We probably need to restructure this so both the icon and this div that contains the TE can be clicked.  The easiest is probably just moving this div inside of the <a> and then tweaking the CSS.

mklewandowski: image

js/core/transform-json-response.ts
```diff
@@ -44,6 +44,7 @@ export interface IPortalData {
     activityUrl: string;
     id: number;
     teacher: string;
+    hasTeacherEdition: boolean;
```
mklewandowski, line 47: How is this value determined?  Is this on a per activity basis?  So we're not going to always show the teacher edition button, but only show it if think an activity has teacher edition content? Am I understanding this correctly?

mklewandowski (changes requested):
Looks good.  My only request is to tweak the teacher edition button so that both areas of the button respond to the click and launch the TE link.
Also, is there a way to see the links working in action (since the mock data just launches broken docker links)?  What's the best way to test with real links?

js/reducers/report-reducer.ts
```diff
@@ -121,7 +124,7 @@ export default function report(state = new ReportState({}), action?: any) {
   switch (action.type) {
     case RECEIVE_PORTAL_DATA:
       data = preprocessPortalDataJSON(action.response);
-
+
```
mklewandowski, line 127: Note that this now throws a lint warning:
```
D:\Repos\portal-report\js\reducers\report-reducer.ts
  127:1  warning  Trailing spaces not allowed  no-trailing-spaces
```
We'll want to fix this.  Might be best to change your VS Code settings to strip whitespace.

js/components/portal-dashboard/question-area.tsx
eireland: done

js/core/transform-json-response.ts
eireland: This is passed in through the offering data the portal receives. So I added it to the mock data.

image

eireland:
To see this in action with the teacher edition, add this activity to a class https://learn.concord.org/eresources/1495.run_resource_html
In the assignments page, show details for the assignment, and open the report.
If activity has no teacher edition, the button should not be shown.

js/reducers/report-reducer.ts
eireland: Fixed

js/core/transform-json-response.ts
eireland: Kiley said button should only be shown when there is teacher edition is available.

mklewandowski:
> To see this in action with the teacher edition, add this activity to a class https://learn.concord.org/eresources/1495.run_resource_html
> In the assignments page, show details for the assignment, and open the report.
> If activity has no teacher edition, the button should not be shown.

Maybe an overly simple question, but is there a URL param that I set to point the report to a specific activity?  Can I point it directly at https://learn.concord.org/eresources/1495.run_resource_html only using URL params?  Or do I have to do this in code?

mklewandowski (approved):
👍 LGTM!

PR #231, opened 2020-09-18
UI updates to adhere to spec

mklewandowski:
This PR includes a series of small UI fixes to adhere to the spec:
- Improve the question overlay animation when the component is displayed (both the height and width should appear to grow)
- Update the cover height so we do not hide the legend border
- Make assignment text in header bold
- Fix the background color of the student response container so it lines up with the upper left nav inside of the all student responses view:
image


eireland (approved):
merge it!

PR #233, opened 2020-09-21
Animate student response popup

mklewandowski:
This PR adds a slide in animation to the `StudentResponsePopup` component.  Animations are done using `react-transition-group`.  Note that this is a more simplistic approach than what was used in the previously opened PR seen here:
https://github.com/concord-consortium/portal-report/pull/232


kswenson (approved):
👍 Looks good

PR #235, opened 2020-09-22
Shows fill in the blank and scaffolded answer text open by default bu…

eireland:
…t still need to click on View Work for all other interactives


js/components/portal-dashboard/answers/iframe-answer.tsx
```diff
@@ -0,0 +1,136 @@
...
+  render() {
+    const { showFullAnswer, answer, responsive } = this.props;
+    const answerText = answer.get("answerText");
+
+    return (
+      <div className={`iframe-answer ${responsive ? "responsive" : ""}`} data-cy="iframe-answer">
+        <div className={`iframe-answer-header ${responsive ? "responsive" : ""}`}>
+          { answerText && <div>{ answerText }</div> }
+          {/* This assumes only scaffolded questions and fill in the blank questions have answerTexts */}
+          { (showFullAnswer && !answerText)? this.renderLink() : ""}
```
mklewandowski, line 130: I don't think we want to make an entire copy of 'iframe-answer.js` just to change this one line of code.  When we first investigated this a month or so ago, I was under the impression that we might need to rewrite some of these answer components for the new dash.  We did need to do this for the image answer - it differed significantly and had little to no internal logic outside of the render.  However, for this component, I think it is ok to continue to use the `.js` version instead of adding an extra `.tsx` version.  You'll see that when I was getting the interactive answers to display better inside of our containers, I added a new prop, `responsive` to `iframe-answer.js` to facilitate the very minor changes that I needed to make.  I think you can do the same.  I would continue to use `iframe-answer.js` and add a new prop (which you did already here) to make this one line tweak to the render in this component.  Then we don't have two components that are over 100 lines but that differ by a line or two.

mklewandowski (changes requested):
Instead of making a duplicate `IframeAnswer` component with a couple lines worth of changes, I think it is better to follow my model on this one and simply make modifications to the existing component (add a prop if needed - although is it true that what you need can be determined from the info already in the component?  Is the `showFullAnswer` prop already being passed to the component?).  It looks like all you need to do is tweak a line or two inside the component.  Anyway, I think we want to avoid this much code that is duplicated.  I'm surprised Code Climate didn't melt!

js/components/report/iframe-answer.js
```diff
@@ -112,8 +113,9 @@ export default class IframeAnswer extends PureComponent {
     return (
       <div className={`iframe-answer ${responsive ? "responsive" : ""}`} data-cy="iframe-answer">
         <div className={`iframe-answer-header ${responsive ? "responsive" : ""}`}>
-          { answerText && <div>{ answerText }</div> }
-          { !alwaysOpen && this.renderLink() }
+          { answerText && <div>{ renderHTML(answerText) }</div> }
+          {/* This assumes only scaffolded questions and fill in the blank questions have answerTexts */}
+          { (!answerText && (!alwaysOpen && this.renderLink())) }
```
mklewandowski, line 118: Does this change the behavior on the previous uses of this component?  Before we had:
`{ !alwaysOpen && this.renderLink() }`
Now we have:
`{ (!answerText && (!alwaysOpen && this.renderLink())) }`

mklewandowski: Ah, I think I get it now.  Doesn't seem to affect existing answers.  You might want to refactor this to be a little clearer with:
```
          { answerText ?
            <div>{ renderHTML(answerText) }</div>
            : !alwaysOpen && this.renderLink() /* This assumes only scaffolded questions and fill in the blank questions have answerTexts */
          }
```

mklewandowski (approved):
👍 I added a suggestion for refactoring one chunk to be clearer.  Otherwise looks good!

js/components/report/iframe-answer.js
eireland: That's what I did at first but then realized that the mock data that was originally there had and answerText for the interactives (cypress tests started failing for the old reports), so I put it back in and made the logic to whether or not display the link separate.

mklewandowski: Huh, I would think it would do the same thing.

mklewandowski: We have `answerText` and `!answerText` - that should work in a ternary.

mklewandowski: Unless it's a subtlety about JS negation that I'm missing.

mklewandowski: If it truly makes the tests fail to add my code, then I would leave it.  I would at least double check it.

js/components/report/iframe-answer.js
```diff
@@ -112,8 +113,10 @@ export default class IframeAnswer extends PureComponent {
     return (
       <div className={`iframe-answer ${responsive ? "responsive" : ""}`} data-cy="iframe-answer">
         <div className={`iframe-answer-header ${responsive ? "responsive" : ""}`}>
-          { answerText && <div>{ answerText }</div> }
-          { !alwaysOpen && this.renderLink() }
+        { answerText
+            ? <div>{ renderHTML(answerText) }</div>
+            : !alwaysOpen && this.renderLink() /* This assumes only scaffolded questions and fill in the blank questions have answerTexts */
+        }
```
mklewandowski, line 119: Tab this in.

mklewandowski (approved):
👍 Tab that one section in and merge it!

PR #237, opened 2020-09-22
Fix student name in modal popup

mklewandowski:
This PR fixes the missing student name in the image answer modal dialog when launched from the all student responses window.


eireland (approved):
Merge it!

PR #239, opened 2020-09-23
Improve handling of spotlight dialogs

mklewandowski:
Adds more sophisticated state management of the spotlight dialog.  We previously assumed that if the number of selected students was 0, that we could always show the spotlight message dialog.  This is incorrect, a user can remove all of the selected students inside the spotlight list dialog itself which then keeps the spotlight list dialog up but with no student showing.  I wasn't eager to introduce another state variable, but I think this is the most straightforward way to handle this other than merging the dialogs into a single component.

### DIFF
--- a/js/components/portal-dashboard/all-responses-popup/student-responses-popup.tsx
+++ b/js/components/portal-dashboard/all-responses-popup/student-responses-popup.tsx
@@ -56,7 +56,7 @@ export class StudentResponsePopup extends React.PureComponent<IProps, IState> {
             setAnonymous={setAnonymous}
             setStudentFilter={setStudentFilter}
             trackEvent={trackEvent}
-            onShowDialog={selectedStudents.length > 0 ? this.setShowSpotlightListDialog(true) : this.setShowSpotlightDialog(true)}
+            onShowDialog={selectedStudents.length > 0 ? this.setShowSpotlightListDialog : this.setShowSpotlightDialog}
           />
           <div className={`${css.questionArea} ${css.column}`} data-cy="questionArea">
             <QuestionNavigator
@@ -83,7 +83,7 @@ export class StudentResponsePopup extends React.PureComponent<IProps, IState> {
             currentActivity={currentActivity}
             currentQuestion={currentQuestion}
             isAnonymous={isAnonymous}
-            onCloseDialog={this.setShowSpotlightListDialog(false)}
+            onCloseDialog={this.setShowSpotlightListDialog}
             onStudentSelect={this.toggleSelectedStudent}
             selectedStudents={selectedStudents}
             setAnonymous={setAnonymous}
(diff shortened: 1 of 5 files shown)


js/components/portal-dashboard/all-responses-popup/student-responses-popup.tsx
```diff
@@ -29,32 +29,34 @@ interface IProps {
...
           <PopupClassNav
             anonymous={anonymous}
-            isSpotlightOn={showSpotlightDialog && selectedStudents.length > 0}
+            isSpotlightOn={selectedStudents.length > 0}
             studentCount={studentCount}
             setAnonymous={setAnonymous}
             setStudentFilter={setStudentFilter}
             trackEvent={trackEvent}
-            onShowDialog={this.showSpotlightDialog(true)}
+            onShowDialog={selectedStudents.length > 0 ? this.setShowSpotlightListDialog(true) : this.setShowSpotlightDialog(true)}
```
kswenson, line 59: Calling `setState()` in render is an anti-pattern. Ideally, you should set the state outside of any render-related function, but if necessary you can do so in `componentDidUpdate()` with a test so that you don't call `setState()` when the value hasn't changed. Otherwise, there's potential for infinite render loops.

kswenson (changes requested):
Needs work on when/where to update the state.

js/components/portal-dashboard/all-responses-popup/student-responses-popup.tsx
mklewandowski: Hmm... I think this is a little deceptive.  With the function currying used on `setShowSpotlightListDialog` and `setShowSpotlightDialog` we're not actually setting state here, but instead using the ternary to selectively pass one of two callback functions into `PopupClassNav` that can be used when needed by the component (and the currying let's us specify an argument).  Then again, I might be missing something.  However, I'm not sure that the currying is needed anymore.  This work was massively refactored last night, so I'm not sure that we need this extra layer of complexity anymore.

kswenson: You're absolutely right -- I missed the currying completely. That said, it looks like you never pass the false version of the curried function, so it seems like you could eliminate the boolean argument and just pass the function.

mklewandowski: I pushed a commit that I think should make this clearer.  It removed the function currying on `setShowSpotlightListDialog` an `setShowSpotlightDialog` so that it is more clear what is happening with these functions when we pass them to `PopupClassNav` via props.  Again, I don't believe that either of these cases involved a state change in the render function, but perhaps I'm missing something.

mklewandowski: Also, @kswenson is it ok if I add you as reviewer for future Portal Dash PRs?  We're slim on reviewers on this project, and some of these PRs could use an extra set of eyes.

kswenson (approved):
👍 I think this is fine and I agree that my original concern was unfounded. As a matter of personal preference I might have gone with four functions (`[show|hide]SpotlightDialog` and `[show|hide]SpotlightListDialog`) and then eliminated all the boolean parameters. I suspect the code would seem simpler to the casual reader, but I don't feel strongly about it.

js/components/portal-dashboard/all-responses-popup/student-responses-popup.tsx
kswenson: Sure, I've generally been reading through them anyway. 😉

mklewandowski: The false version was being passed, but to a different component.  Regardless, I think it's a little easier to read in its present form.

mklewandowski: The `PopupClassNav` only has the job of showing the components, whereas the components themselves contain the x close button and have the job of closing themselves.

PR #240, opened 2020-09-24
Update stored spotlight state to use student ids and not student map

mklewandowski:
This PR changes how we store the currently selected set of spotlight students in the `StudentResponsePopup` component state.  Previously we stored  an array of students (where students was an immutable JS map, the same way that students are stored in the Redux state).  These changes make it so that we now store an array of student id strings in the component state.

It wasn't immediately obvious why this change was needed, but there was a bug in the anonymized name that appeared in the student spotlight.  This made it clear that our previous component state model was incorrect.  Previously, when a student was selected for spotlight, we grabbed a copy of the student (again, an immutable JS map) from the Redux state and stored that in our component state.  However, after the student state in Redux is initially built, it does NOT remain static.  Anonymizing the list of students, actually changes the students in the Redux state (student name is updated).  However, if we've already built a local array of students in the `StudentResponsePopup` component, we don't have access to any student state changes on the Redux state.  The safer approach is to just have the `StudentResponsePopup` store student ids in component state.  The ids can be used to access students from the Redux state.  Then if properties of the student change in the Redux state (like when we anonymize the names), we can use our student id to access a student out of the Redux state and ensure that our set of selected spotlight students is always up to date (instead of having some potentially obsolete copy).

### DIFF
--- a/js/components/portal-dashboard/all-responses-popup/popup-student-response-list.tsx
+++ b/js/components/portal-dashboard/all-responses-popup/popup-student-response-list.tsx
@@ -8,22 +8,22 @@ import css from "../../../../css/portal-dashboard/all-responses-popup/popup-stud
 interface IProps {
   currentQuestion?: Map<string, any>;
   isAnonymous: boolean;
-  onStudentSelect: (student: Map<any, any>) => void;
-  selectedStudents: Map<any, any>[];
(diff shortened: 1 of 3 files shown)


kswenson (approved):
👍 Looks good

PR #241, opened 2020-09-24
174915493 student sort

mklewandowski:
This PR syncs the state of the two student sort `CustomSelect` component instances within the app.  Unlike other occurrences of the `CustomSelect` where the selected item is managed by the component's local state, if the current sort method is changed in one `CustomSelect`, it must also be changed in the other instance (this is similar to what we needed to do with the 3(!) anonymize toggles in the app - they all have to have a synchronized state).  To do this, we obtain the current student sort method from the Redux component and pass it to the `CustomSelect`.  This allows the `CustomSelect` to use or override its internal component state management based on the absence or presence of an optional prop.

There is also some minor refactoring of component props lists.  As the props lists grow on these components (a side effect of an early choice to grab the Redux state in the main app component and then pass it down as needed to child components), I'm finding it helpful to have them in alphabetical order.

### DIFF
--- a/js/components/portal-dashboard/custom-select.tsx
+++ b/js/components/portal-dashboard/custom-select.tsx
@@ -6,17 +6,18 @@ import { SvgIcon } from "../../util/svg-icon";
 import css from "../../../css/portal-dashboard/custom-select.less";
 
 interface IProps {
+  dataCy: string;
+  disableDropdown?: boolean;
+  HeaderIcon: SvgIcon;
+  isHeader?: boolean;
   items: SelectItem[];
   onSelectItem: (value: string) => void;
+  selectState?: string;
   trackEvent: (category: string, action: string, label: string) => void;
-  HeaderIcon: SvgIcon;
-  dataCy: string;
-  isHeader?: boolean;
-  disableDropdown?: boolean;
 }
 
 interface IState {
-  current: string;
+  selected: string;
   showList: boolean;
 }
 
@@ -30,7 +31,7 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
   constructor(props: IProps) {
     super(props);
     this.state = {
-      current: props.items[0].action,
+      selected: props.items[0].action,
       showList: false
     };
   }
@@ -53,8 +54,9 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
   }
 
   private renderHeader = () => {
-    const { items, HeaderIcon } = this.props;
-    const currentItem = items.find(i => i.action === this.state.current);
+    const { items, HeaderIcon, selectState } = this.props;
+    const selectedItem = selectState ? selectState : this.state.selected;
+    const currentItem = items.find(i => i.action === selectedItem);
     const showListClass = this.state.showList ? css.showList : "";
     const useHeader = this.props.isHeader ? css.topHeader : "";
     const disabled = this.props.disableDropdown ? css.disabled : "";
@@ -68,12 +70,13 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
   }
 
   private renderList = () => {
-    const { items } = this.props;
+    const { items, selectState } = this.props;
+    const selectedItem = selectState ? selectState : this.state.selected;
     const useHeader = this.props.isHeader ? css.topHeader : "";
     return (
       <div className={`${css.list} ${useHeader} ${(this.state.showList ? css.show : "")}`}>
         { items && items.map((item: SelectItem, i: number) => {
-          const currentClass = this.state.current === item.action ? css.selected : "";
+          const currentClass = selectedItem === item.action ? css.selected : "";
           return (
             <div
               key={`item ${i}`}
@@ -104,11 +107,11 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
     });
   }
 
-  private handleListClick = (current: string) => () => {
-    this.props.onSelectItem(current);
-    this.props.trackEvent("Portal-Dashboard", "Dropdown", current);
+  private handleListClick = (selected: string) => () => {
+    this.props.onSelectItem(selected);
+    this.props.trackEvent("Portal-Dashboard", "Dropdown", selected);
     this.setState({
-      current,
+      selected,
       showList: false
     });
   }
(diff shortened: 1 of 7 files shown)


js/components/portal-dashboard/custom-select.tsx
```diff
@@ -6,17 +6,18 @@ import { SvgIcon } from "../../util/svg-icon";
 import css from "../../../css/portal-dashboard/custom-select.less";
 
 interface IProps {
+  dataCy: string;
+  disableDropdown?: boolean;
+  HeaderIcon: SvgIcon;
+  isHeader?: boolean;
   items: SelectItem[];
   onSelectItem: (value: string) => void;
+  selectState?: string;
```
kswenson, line 15: As suggested below, I think the naming could be clearer here. If I were just naming this prop I'd probably name it `selectedItem` for consistency with `onSelectItem`, but that's not ideal because it's not really an item. If we think of this as a simple control like any other, e.g. a standard `<select>`, then the props should be named `value` and `onChange`. Then perhaps the fields of `SelectItem` should be named `value` and `label`. More broadly, I think there was a notion that `CustomSelect` was a component that could be standardized for use across multiple CC applications, but instead each usage has been customized in unique ways so that much of the benefit of consistency has been lost. Ideally, any changes to this component should be made with an eye toward general applicability rather than local expediency. I'm not sure how much of this is worth addressing now, but it's something to think about.

js/components/portal-dashboard/custom-select.tsx
```diff
@@ -53,8 +54,9 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
   }
 
   private renderHeader = () => {
-    const { items, HeaderIcon } = this.props;
-    const currentItem = items.find(i => i.action === this.state.current);
+    const { items, HeaderIcon, selectState } = this.props;
+    const selectedItem = selectState ? selectState : this.state.selected;
+    const currentItem = items.find(i => i.action === selectedItem);
```
kswenson, line 59: I find the naming here somewhat confusing. `selectState` is a prop not state. `selectedItem` appears to be an action not an item. The `currentItem` is presumably also the selected item.

js/components/portal-dashboard/custom-select.tsx
```diff
@@ -68,12 +70,13 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
   }
 
   private renderList = () => {
-    const { items } = this.props;
+    const { items, selectState } = this.props;
+    const selectedItem = selectState ? selectState : this.state.selected;
```
kswenson, line 74:
```suggestion
    const selectedItem = selectState || this.state.selected;
```

js/components/portal-dashboard/custom-select.tsx
```diff
@@ -53,8 +54,9 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
   }
 
   private renderHeader = () => {
-    const { items, HeaderIcon } = this.props;
-    const currentItem = items.find(i => i.action === this.state.current);
+    const { items, HeaderIcon, selectState } = this.props;
+    const selectedItem = selectState ? selectState : this.state.selected;
```
kswenson, line 58:
```suggestion
    const selectedItem = selectState || this.state.selected;
```

kswenson (approved):
👍 I have some misgivings about the direction that these changes are taking the `CustomSelect` component but I don't know that there's an easy way to address them so I leave it to you to decide to what extent changes should be made now rather than put off to later.

js/components/portal-dashboard/custom-select.tsx
```diff
@@ -30,7 +31,7 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
   constructor(props: IProps) {
     super(props);
     this.state = {
-      current: props.items[0].action,
+      selected: props.items[0].action,
```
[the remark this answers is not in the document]
mklewandowski: I think you're right.  We were lucking out that the initial state was correct., but I believe it needs to be derived from either of the props.

js/components/portal-dashboard/custom-select.tsx
mklewandowski: I like these changes.  In making them, I would also vote to turn the state variable `selected` also into `value` so we have consistency with the prop.  I also agree on attempting to make things like the `CustomSelect` and the `ToggleControl` generic and reusable in other apps.  That might be a fun side-project to add them to the CC components library that others were working on:
https://github.com/concord-consortium/concord-react-components

mklewandowski: And in line with your naming ideas, I think we can turn `handleListClick` into `handleChange` for consistency.

js/components/portal-dashboard/custom-select.tsx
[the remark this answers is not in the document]
mklewandowski: Fixed.

js/components/portal-dashboard/custom-select.tsx
mklewandowski: I made the above changes.

js/components/portal-dashboard/custom-select.tsx
mklewandowski: Fixed.

js/components/portal-dashboard/custom-select.tsx
mklewandowski: Fixed.

eireland (approved):
Looks good to me

mklewandowski (commented):
> Looks good to me

I'll wait until @kswenson has a chance to look at my select changes before merging.

js/components/portal-dashboard/custom-select.tsx
```diff
@@ -53,36 +54,38 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
...
+    const currentValue = value || this.state.value;
+    const currentItem = items.find(i => i.value === currentValue);
     const showListClass = this.state.showList ? css.showList : "";
     const useHeader = this.props.isHeader ? css.topHeader : "";
     const disabled = this.props.disableDropdown ? css.disabled : "";
     return (
       <div className={`${css.header} ${useHeader} ${showListClass} ${disabled}`} onClick={this.handleHeaderClick}>
         { <HeaderIcon className={`${css.icon} ${showListClass}`} /> }
-        <div className={css.current}>{currentItem && currentItem.name}</div>
+        <div className={css.current}>{currentItem && currentItem.label}</div>
```
kswenson, line 66:
```suggestion
        <div className={css.current}>{currentItem?.label}</div>
```

kswenson (approved):
👍 Looks good

js/components/portal-dashboard/custom-select.tsx
mklewandowski: Fixed.

PR #242, opened 2020-09-24
Fix crash in spotlight

mklewandowski:
This PR fixes two bugs in the spotlight view:

- The app crashed when opening an interactive question in the spotlight.  This was caused by an empty prompt in the question.  I expanded the checks for what text to display for the question in the spotlight, but @eireland you did much of the work that decides which question text to display in which context.  You might want to confirm my text choices.
- The vertical size of the content container in the spotlight view was incorrect.  It didn't fill the entire space of the spotlight view.  As a side note, I would avoid using the CSS `calc` with the viewport height/width (`vh` and `vw`) unless absolutely needed.  It tends to lock us into complex calculations that depend upon other component sizing when simple flex containers can often do the work (and then you often don't need to worry about future changes to the calculation).

image

### DIFF
--- a/js/components/portal-dashboard/all-responses-popup/spotlight-student-list-dialog.tsx
+++ b/js/components/portal-dashboard/all-responses-popup/spotlight-student-list-dialog.tsx
@@ -61,7 +61,7 @@ export class SpotlightStudentListDialog extends React.PureComponent<IProps>{
     const type = currentQuestion?.get("type");
     const questionType = QuestionTypes.find(qt => qt.type === type);
     const QuestionIcon = questionType?.icon;
-    const typeText = type && type.replace(/_/gm, ' ');
+    const typeText = type?.replace(/_/gm, ' ');
     const interactiveName = type === "iframe_interactive" ? currentQuestion?.get("name") : typeText;
     const prompt = currentQuestion?.get("prompt");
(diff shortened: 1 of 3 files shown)


js/components/portal-dashboard/all-responses-popup/spotlight-student-list-dialog.tsx
```diff
@@ -61,7 +61,10 @@ export class SpotlightStudentListDialog extends React.PureComponent<IProps>{
     const type = currentQuestion?.get("type");
     const questionType = QuestionTypes.find(qt => qt.type === type);
     const QuestionIcon = questionType?.icon;
-    const prompt = currentQuestion ? striptags((currentQuestion.get("prompt")).replace(/&nbsp;/g, ' ')) : "";
+    const typeText = type && type.replace(/_/gm, ' ');
```
kswenson, line 64:
```suggestion
    const typeText = type?.replace(/_/gm, ' ');
```

kswenson (approved):
👍

js/components/portal-dashboard/all-responses-popup/spotlight-student-list-dialog.tsx
mklewandowski: Fixed.

mklewandowski:
@eireland what should we do if the prompt is very large?  Like the fill in the blank question?

PR #243, opened 2020-09-24
Fix multiple choice question layout

mklewandowski:
This PR fixes a bug that caused the multiple choice bubble to display incorrectly when we have long text in the multiple choice question.  Changes include:
- fix bug so bubble displays properly
- reduce margin so we have more space to display question
- remove undefined `mcContent` class that we were adding to the question text

image


kswenson (approved):
👍

PR #244, opened 2020-09-24
Toggle current activity and close question overlay when click activity in level viewer

mklewandowski:
This PR makes changes so that after clicking on an activity button in the Level Viewer, we clear the current selected question and hide the Question Overlay.  This adheres to the Dash Prototype (https://hav7ta.axshare.com/#id=nkd8s4&p=portal_dashboard___reports_v3_4), and fixes some of the odd states that a user can get into in the app.  Without this fix, the following can happen:
- User has activity1: question1 showing in Question Overlay.  User clicks on activity2 in the Level Viewer.  We still display activity1: question1 in the Question Overlay even though the user has selected activity2.  
image

- User has Question Overlay showing.  User clicks activity button in Level Viewer to collapse the current activity (now NO activity is selected).  User now presses the Show All Students buttons.  Window appears but there is no activity title shown, because no activity is selected. @eireland we saw this the other day while pairing.
image

### DIFF
--- a/js/components/portal-dashboard/level-viewer.tsx
+++ b/js/components/portal-dashboard/level-viewer.tsx
@@ -189,6 +189,7 @@ export class LevelViewer extends React.PureComponent<IProps> {
   }
 
   private handleActivityButtonClick = (activityId: string) => () => {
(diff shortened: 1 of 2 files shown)


kswenson (approved):
👍 LGTM

PR #245, opened 2020-09-25
174962304 text overflow

mklewandowski:
This PR fixes two cases where long, overflowing text was not being handled properly.  In both cases overflow text should be shown with a set number of lines and with ellipsis as per the UI spec.  Cases include:
- activity names in the buttons in the level viewer
- activity name in the all student responses view
image

For anyone interested, I used the library `react-lines-ellipsis` to handle multi-line overflow (which is much trickier than single line overflow with ellipsis).  I also tried `react-clamp-lines` (I couldn't get the sizing and number of clamped lines to work) and `React-clamp` (old, didn't work at all).

### DIFF
--- a/css/portal-dashboard/level-viewer.less
+++ b/css/portal-dashboard/level-viewer.less
@@ -64,7 +64,7 @@
         cursor: pointer;
 
         .activityTitle {
-          width: 73px;
+          width: 100px;
           height: 50%;
           padding: 6px 10px 0 10px;
           text-overflow: ellipsis;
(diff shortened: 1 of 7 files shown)


eireland (approved):
Looks good. I'm wondering if we should also implement this in the page titles 
image

mklewandowski:
> Looks good. I'm wondering if we should also implement this in the page titles
> image

Ah, yes!  Good catch!  I added the ellipsis to the page name container.  I also added tooltip text so that when a user hovers, they can see the full text.

kswenson (commented):
👍

PR #246, opened 2020-09-25
Reset spotlighted students when change question

mklewandowski:
As per the design discussion, this PR clears the set of spotlighted students when the user navigates to a new question in the All Student Responses view.


eireland (approved):
merge it!

kswenson (commented):
👍

PR #250, opened 2020-09-26
Add dynamic spotlight colors

mklewandowski:
This PR adds highlight colors to the spotlight student rows.  The highlight colors are changed by pressing the spotlight button next to the student name.

Note that I'm using an array of `SelectedStudent` objects in my React state inside the `StudentResponsePopup` component.  I think my state changes are safe, but there is some trickiness here when dealing with changing object properties inside of an array in state.  In my case I'm making a new `SelectedStudent` object that gets put into the array - so I think I'm safe.  But this is worth looking at.

spotlightcolors

### DIFF
--- a/css/portal-dashboard/all-responses-popup/spotlight-student-list-dialog.less
+++ b/css/portal-dashboard/all-responses-popup/spotlight-student-list-dialog.less
@@ -206,9 +206,6 @@
             object-fit: contain;
           }
 
-          &:hover .listSpotlightIcon {
-            fill: white;
-          }
           &:active .listSpotlightIcon {
             fill: white;
           }
--- a/css/portal-dashboard/all-responses-popup/spotlight-student-list-dialog.less
+++ b/css/portal-dashboard/all-responses-popup/spotlight-student-list-dialog.less
@@ -193,10 +193,24 @@
           background-color: white;
           margin-left: 10px;
           box-sizing: content-box;
+          cursor: pointer;
+
+          &:hover {
+            background-color: @cc-teal-light3;
+          }
+          &:active {
+            background-color: @cc-teal;
+          }
 
           .listSpotlightIcon {
             object-fit: contain;
-            fill: @cc-teal;
+          }
+
+          &:hover .listSpotlightIcon {
+            fill: white;
+          }
+          &:active .listSpotlightIcon {
+            fill: white;
           }
         }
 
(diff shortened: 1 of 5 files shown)


eireland (approved):
just the one color change. Otherwise looks good

css/portal-dashboard/all-responses-popup/spotlight-student-list-dialog.less
```diff
@@ -193,10 +193,24 @@
...
+            background-color: @cc-teal;
+          }
 
           .listSpotlightIcon {
             object-fit: contain;
-            fill: @cc-teal;
+          }
+
+          &:hover .listSpotlightIcon {
+            fill: white;
```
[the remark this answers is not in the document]
mklewandowski: Ah, it is wrong.  From studying the prototype, it looks like the icon does not change color when hovered (just the background).  This needs to come out entirely.  The icon only turns white when it is active.  Fixed.

js/components/portal-dashboard/all-responses-popup/student-responses-popup.tsx
```diff
@@ -103,9 +109,23 @@ export class StudentResponsePopup extends React.PureComponent<IProps, IState> {
     );
   }
 
+  private handleChangeColorGroup = (studentId: string) => {
+    const { selectedStudents } = this.state;
+    const index = selectedStudents.findIndex((s: SelectedStudent) => s.id === studentId);
+    if (index >= 0) {
```
kswenson, line 115: As an aside I prefer this formulation ("index is valid") to the seemingly more common `index !== -1` which always reads to me as "result is not equal to the magic sentinel value for failure" which seems like an odd way to test for success.

js/components/portal-dashboard/all-responses-popup/student-responses-popup.tsx
```diff
@@ -103,9 +109,23 @@ export class StudentResponsePopup extends React.PureComponent<IProps, IState> {
...
+      const currentColor = selectedStudents[index].colorGroup;
+      const newColor = currentColor >= spotlightColors.length - 1 ? 0 : currentColor + 1;
+      const updatedStudent: SelectedStudent = { id: studentId, colorGroup: newColor };
+      const updatedSelectedStudents = [...selectedStudents];
+      updatedSelectedStudents.splice(index, 1);
+      updatedSelectedStudents.push(updatedStudent);
```
kswenson, line 121: As long as we're dealing with simple JavaScript objects (e.g. not immutable objects), you don't actually need to splice in a new student object. Another approach in this case, by the way, would be a map which passed most objects through unchanged and only changed the one of interest, e.g. something like:
```
  const updatedSelectedStudents = selectedStudents.map(s => {
    return (s.id === studentId) ? { id: studentId, colorGroup: newColor } : s;
  });

kswenson (approved):
👍 Looks good -- I suggested a simplification in one place.

PR #251, opened 2020-09-28
Improve level viewer animation

mklewandowski:
Improve the animation of level viewer contents by putting the level viewer behind other components.

animate


eireland (approved):
merge it!

PR #252, opened 2020-09-28
Fix question area height in question overlay and all student responses

mklewandowski:
Improve handling of long question text in both the question overlay and all student response views.

image

### DIFF
--- a/css/portal-dashboard/question-area.less
+++ b/css/portal-dashboard/question-area.less
@@ -2,8 +2,6 @@
 
 .questionArea {
   flex-basis: auto;
-  overflow-y: auto;
-  max-height: calc(100vh * .4);
 }
 
 .questionTypeHeader {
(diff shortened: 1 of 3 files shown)


css/portal-dashboard/question-area.less
```diff
@@ -2,6 +2,8 @@
 
 .questionArea {
   flex-basis: auto;
+  overflow-y: auto;
```
eireland, line 5: I'm wondering if this would be better in the .questionTextArea instead so we keep the question type, activity link, and teacher edition button in view, and we only scrolll the actual question text?

mklewandowski: Yeah, that does feel a little better with the header area that has the question type and links remaining at the top.

mklewandowski: A pushed a change that contains your suggestion.  I think it's better now.

eireland (approved):
merge it!

PR #255, opened 2020-11-03
172894510 anonymous student report

eireland:
Adds ability to open an anonymous student's report from the Generate Report button at the end of an activity. There was some work aslo done in activity-player, and report-services repos

PR #256, opened 2020-11-13
Add anonymous user to be able to open portal report from Activity Player

eireland:
Fixes initial data state fetching error to show in site instead of just showing a spinner
Adds ability for anonymous user to launch a portal report at the end of an activity in the activity player
Adding runKey query param for anonymous reports


pjanik (changes requested):
I think I reviewed almost identical PR and I left comments there that are still applicable here: https://github.com/concord-consortium/portal-report/pull/255

js/actions/index.ts
```diff
@@ -74,7 +93,12 @@ function _receivePortalData(db: firebase.firestore.Firestore,
     type: RECEIVE_PORTAL_DATA,
     response: rawPortalData
   });
-  let resourceUrl = rawPortalData.offering.activity_url.toLowerCase();
+  let resourceUrl;
+  if ((rawPortalData.offering.activity_url).includes("activity=")) {
```
scytacki, line 97: This is a kind of hack to support reporting on activity player based external activities. In those cases the url will look something like: `https://activity-player.concord.org?activity=https%3A%2F%2Fauthoring.concord.org%2Factivity%2F123`

A comment to this effect should be added to this code.

Also the parsing of the URL is fragile. It'd be better to use the queryString library to parse the url and pull out the activity parameter.

js/actions/index.ts
```diff
@@ -243,6 +266,16 @@ function watchCollection(db: firebase.firestore.Firestore, path: string, receive
     snapshot => snapshot.docs.map(doc => doc.data()));
 }
 
+function watchAnonymousAnswers(db: firebase.firestore.Firestore, source: string, runKey: string, dispatch: Dispatch) {
+// Setup Firebase observer. It will fire each time the resource structure is updated.
```
scytacki, line 270: This comment doesn't match what the function is doing. It will fire when the answers are updated.

js/db.ts
```diff
@@ -57,6 +57,7 @@ async function initializeDB(name: string) {
 
   if (urlParam("enableFirestorePersistence")) {
     await firebase.firestore().enablePersistence({ synchronizeTabs: true });
+    await firebase.firestore().disableNetwork();
```
scytacki, line 60: This seems like a strange thing be adding in this PR.  I don't remember exactly but I think the cypress tests have another way of disabling the network.  So I think this change should be removed unless it can be justified.

js/reducers/data-reducer.ts
```diff
@@ -11,7 +11,8 @@ export interface IDataState {
 }
 
 const INITIAL_DATA_STATE = RecordFactory<IDataState>({
-  isFetching: true
+  isFetching: true,
+  error: null
```
scytacki, line 15: Is this new error property used anywhere? It doesn't seem to be.

scytacki (changes requested):
I found a few things to fix or clarify.

README.md
```diff
@@ -123,17 +123,49 @@ replace the url host and path with `localhost:8080`. Alternatively, if you are a
...
 when you view the class details, you will see a "Local Tracked Q" button next to "Report" which will link to localhost.
 
-### Additional URL Parameters
+### URL Parameters
+
+* `token={string}`: access token provided by the portal when it launches the report. It is used as a Bearer
+                    token when making requests to the portal
+* `class={url}`:    URL to get info about the class from the portal
+* `offering={url}`: URL to get info about the offering from the portal
+* `firebase-app={name}`: identifier for a portal Firebase App. This name is send to the portal to get a FirebaseJWT.
```
kswenson, line 132:
```suggestion
* `firebase-app={name}`: identifier for a portal Firebase App. This name is sent to the portal to get a FirebaseJWT.
```

README.md
```diff
@@ -123,17 +123,49 @@ replace the url host and path with `localhost:8080`. Alternatively, if you are a
...
+                    It defaults to "report-service".
+* `tool-id={uri}`:  URI identifying the tool that stored the answers in Firestore. The tool-id is converted to a
+                    source key, which is used to query the answers in firestore. The conversion from tool-id to
+                    source key matches the make_source_key method in LARA's report_service.rb
+* `portal-dashboard`: boolean parameter which tells the report to render in a new dashboard style
+* `dashboard`:      boolean parameter which tells the report to render in old dashboard style
+* `activityIndex={index}`: when the activity is a sequence, only show this activities questions
```
kswenson, line 139:
```suggestion
* `activityIndex={index}`: when the activity is a sequence, only show this activity's questions
```

README.md
```diff
@@ -123,17 +123,49 @@ replace the url host and path with `localhost:8080`. Alternatively, if you are a
...
+* `studentId={id}`: This shows the report for a single student, and removes some UI affordances. The filtering of the
+                    student data happens client-side.
+* `iframeQuestionId={id}`: This, combined with a valid `studentId`, will show a stand-alone, full-size iframe containing
+                    the model referenced by iframeQuestionId, and the answer saved by studentId (either as state or as
```
kswenson, line 143:
```suggestion
                    the model referenced by `iframeQuestionId`, and the answer saved by `studentId` (either as state or as
```

README.md
```diff
@@ -123,17 +123,49 @@ replace the url host and path with `localhost:8080`. Alternatively, if you are a
...
-* `studentId={id}`:   This shows the report for a single student, and removes some UI affordances. The filtering of the
-                      student data happens client-side.
-* `iframeQuestionId={id}`: This, combined with a valid `studentId`, will show a stand-alone, full-size iframe containing
-                      the model referenced by iframeQuestionId, and the answer saved by studentId (either as state or as
-                      a url).
 * `enableFirestorePersistence=true`: Uses a local firestore DB for data persistance across sessions and tabs. Clear the
-                      DB by going to `dev tools > Application > IndexedDB > firebaseLocalStorageDb > Delete database`
+                    DB by going to `dev tools > Application > IndexedDB > firebaseLocalStorageDb > Delete database`
+* `clearFirestorePersistence=true`: Clears local firestore DB. If this and `enableFirestorePersistence=true` are set
+                    then the DB will cleared first before the local persistence is enabled.
```
kswenson, line 168:
```suggestion
                    then the DB will be cleared first before the local persistence is enabled.
```

js/actions/index.ts
scytacki: I think this code can stay as it is. We'll likely be refactoring it when cleaning up the changes needed to run the report directly from the portal.

js/reducers/data-reducer.ts
scytacki: I faintly remember that we added this to quickly fix the error reporting code. It should probably be removed, and if the error reporting code breaks then we should fix that code instead of making the change here.

scytacki:
The changes look good,

You could modify urlParam and urlHashParam to use the urlStringParam function:
```
...
return urlStringParam(window.location.search, name);
...
return urlStringParam(window.location.hash, name);
...
```
That will help with test coverage too since now anything that covers urlParam will also  cover urlStringParameter.

If we had a test with a runKey in the URL that would cover a lot this new stuff, but I'm not sure how easy that would be to setup.

We should try to increase the coverage though before merging this.

scytacki (approved):
I think this is good enough. 

@eireland and I worked on this and there isn't an easy way to cover the anonymous data with tests.  We'd need to add the resource structure and answers to firestore and run it offline to get better coverage.

js/reducers/data-reducer.ts
scytacki: @eireland I think is related to the error we were seeing in cypress when the network was not disabled. Looking at this code more closely I now remember that we added this because the `.set("error",...` later on in this file would fail if there wasn't an error property set on the Object.

I think the best thing to do here is to add a test of this error handling, so we can verify this. In the meantime this PR can just leave the code as it was before, so no change from the current state of this PR.

PR #258, opened 2020-11-20
174046971 logged in user report

scytacki:
This builds on the work of the anonymous-student report branch.

It uses the OAuth to log the user in via the portal, and get an access token. Then this token is used get information about the class, offering, and get a JWT to work with firestore.


eireland:
Fixed the failing cypress tests.

cypress/integration/portal-report-smoke-test.spec.js
```diff
@@ -20,7 +20,9 @@ context("Portal Report Smoke Test", () => {
     // run, so this doesn't seem necessary.
 
     beforeEach(() => {
-        cy.visit("/");
+      let fakeServer = "http://portal.test";
+
+        cy.visit(`/?reportUrl=${encodeURIComponent(fakeServer)}`);
```
scytacki, line 25: The reportUrl param is deprecated in the portal, so it shouldn't be used here in the tests.

js/api.ts
```diff
@@ -154,11 +200,14 @@ function fakeUserType(): "teacher" | "learner" {
...
-  return classPromise.then((classData: any) => {
-    const firestoreJWTPromise = fetchFirestoreJWT(classData.class_hash);
-    return Promise.all([offeringPromise, firestoreJWTPromise]).then((result: any) => {
-      const offeringData = result[0];
-      const rawFirestoreJWT = result[1].token;
+  return Promise.all([offeringPromise, classPromise]).then(([offeringData, classData]: [any, any]) => {
+    const resourceLinkId = offeringData.id.toString();
+    // only pass resourceLinkId if there is a studentId
+    // FIXME: if this is a teacher viewing the report of a student there will be a studentId
+    // but the token will be for a teacher, so then the resourceLinkId should be null
```
scytacki, line 207: Normally I'd say this shouldn't be merged with a `FIXME` but in this case the issue is with another feature which isn't part of this PR: supporting launching of an AP report from the portal.
So it seems fine to leave this in and we'll need to deal with it when implementing that feature.

js/components/report/interactive-iframe.js
```diff
@@ -34,8 +35,14 @@ export default class InteractiveIframe extends PureComponent {
...
 
   handleGetFirebaseJWT = (options) => {
-    return fetchClassData()
-      .then(classData => fetchFirestoreJWT(classData.class_hash, options.firebase_app))
+    return Promise.all([fetchOfferingData(), fetchClassData()])
+      .then(([offeringData, classData]) => {
+        const resourceLinkId = offeringData.id.toString();
+        // only pass resourceLinkId if there is a studentId
+        // FIXME: if this is a teacher viewing the report of a student there will be a studentId
+        // but the token will be for a teacher, so then the resourceLinkId should be null
```
scytacki, line 43: Same as the FIXME above.

scytacki (commented):
I found one issue in the cypress tests. 

Also I think we need to test the OAuth flow better since this will be a key part of the report now, and right now it doesn't have coverage. I'm not sure yet how to test it, but perhaps we can just use spec tests for each of the methods of the flow.

pjanik:
I'm much less in the context that @scytacki is. Skimming through the code, I don't think I can add more here.

scytacki:
This now has passing tests and includes coverage of the OAuth flow.

I think it does have a flaw in how it is figuring out the answerSource. But I think that can be fixed in a new PR which will be needed to handle launching the report directly from the portal.

js/api.ts
```diff
@@ -97,7 +131,16 @@ const gePortalReportAPIUrl = () => {
   return null;
 };
 
-const getAuthHeader = () => `Bearer ${urlParam("token")}`;
+export function getAuthHeader() {
+  const fakeServer = "http://portal.test"; //used for testing
```
eireland, line 135: May want to delete the fakeServer reference since it is no longer used

eireland (approved):

PR #260, opened 2020-11-24
Improve anonymous tests

scytacki:
This changes the 'fake' mode for runKey based testing of the report.
It now tries to load the answer documents for real.  This way the tests exercise more of the code.
It does mean when a runKey is specified, then no answers are shown, however if the point of the test is to work with the answers then it should just leave off the runKey param.

It also moves the disableNetwork into db.ts This seems like a more natural place to put this. 

Additionally, it adds a test for the showing of network errors. This uncovered a bug in the code where the Immutable Record was being initialized without a `error` property. And as far as I understand Immutable Records need to be initialized with all possible properties. It doesn't allow `set(..)` calls unless the properties are defined. So the code was throwing an error while trying to render an error.

### DIFF
--- a/cypress/integration/api-error.spec.js
+++ b/cypress/integration/api-error.spec.js
@@ -1,0 +1,10 @@
+context("Test api error", () => {
+  context("when an invalid offering url is used", () => {
+    before(() => {
+      cy.visit("/?offering=http://example.com");
+    });
+    it('shows an error message to user', () => {
+      cy.contains("network error");
+    });
+  });
+});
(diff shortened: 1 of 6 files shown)


pjanik:
It looks good to me. 👍 

> It does mean when a runKey is specified, then no answers are shown, however if the point of the test is to work with the 
answers then it should just leave off the runKey param.

So, no answers will be loaded, but the test can set and read them, as Firestore works locally, right?
And they'll be persisted or cleared based on the enableFirestorePersistence and clearFirestorePersistence values?

eireland (approved):
Looks good

scytacki:
@pjanik, that is correct. The test could set and read them. Right now we don't have examples of the tests setting answers.

I've been thinking it would be better for the tests (and demo) to set both the answers and resource structure in the offline firestore. Then the code for working with firestore would be the same whether working with fake data or not.

PR #261, opened 2020-12-02
175749596 feedback

mklewandowski:
This PR contains infrastructure changes to facilitate the feedback features that are needed for the 2.0 MVP.  None of these changes result in any changes/updates/modifications to the existing UI or functionality.  All of the changes and additions are simply a starting point for the feedback work.

Changes include:
- remove the now obsolete `Feedback` toggle component
- add `showFeedbackBadges` to the Redux state - this is used to show or hide the feedback badges that appear next to answers in the student rows
-  update the header menu and add a `HeaderMenuItem` component which allows us to have multiple stateful entries in the menu.  This is needed since the user will be able to toggle the feedback badges from the menu.  Note that the code for the feedback toggle entry has been added to the header menu component, but is commented out.  Look in `renderMenuItems` or search for the string "TODO: FEEDBACK" to turn this menu item on when we are ready to use it.
- pass `questionFeedbacks` and `showFeedbackBadges` into the `StudentAnswers` component.  These new props will be used to determine if we should show the feedback badges and which badges to show in the student row.  We will need to expand the render function in `StudentAnswers` to show the badges, but the relevant state is now in the component.
- add `ActivityFeedbackPanel` and `QuestionFeedbackPanel` components which will be used to display the feedback panels in the all student responses.  At present, the props are set up in both of these components so that they have the proper Redux state.  We will still need to build the render function and handler functions to show and update the feedback, but these components at least have the relevant state.


emcelroy (approved):

mklewandowski:
@eireland I'm going to merge this PR, but I think it is still worth reading the notes in it in case you need to work on the feedback in the near future.

PR #262, opened 2020-12-04
Update all responses button in question overlay

mklewandowski:
Update the all responses button in the question overlay to match the UI spec.  Changes includes:
- move all responses button to question detail header
- remove all responses button from footer (entire footer should be removed)
- update colors in all responses button (button state colors should change from teal to orange)
- update related text. "View All Student Responses" becomes "View All Responses" and "Question Detail View" becomes "Question Details".

Screen Shot 2020-12-03 at 9 00 47 PM

### DIFF
--- a/css/portal-dashboard/variables.less
+++ b/css/portal-dashboard/variables.less
@@ -25,6 +25,7 @@
 @cc-orange-light2: #ffc18a;
 @cc-orange-light3: #ffcea1;
 @cc-orange-light4: #ffe6d0;
+@cc-orange-light4B: #ffeddc;
 @cc-orange-light5: #fff2e7;
 @cc-orange-light6: #fff9f3;
 @score-green: #19e549;
--- a/css/portal-dashboard/question-overlay.less
+++ b/css/portal-dashboard/question-overlay.less
@@ -27,6 +27,7 @@
     flex: 0 0 50px;
     display: flex;
     align-items: center;
+    justify-content: space-between;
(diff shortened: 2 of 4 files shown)


css/portal-dashboard/variables.less
```diff
@@ -25,6 +25,7 @@
 @cc-orange-light2: #ffc18a;
 @cc-orange-light3: #ffcea1;
 @cc-orange-light4: #ffe6d0;
+@cc-orange-light4B: #ffeddc;
```
emcelroy, line 28: I'm sure it's fine as is, but I can't help but ask if making it `@cc-orange-light5` and then incrementing the next two variables' numbers accordingly would be problematic?

emcelroy (approved):
Looks good 👍 I just had one comment on something that's probably not important and should very likely stay as is.

css/portal-dashboard/variables.less
mklewandowski: Good question.  I tend to defer to Michael for color names and use whatever naming conventions that he uses in the UI spec.  This creates consistency when matching up colors in the UI spec and the actual application CSS.  In this case, Michael went with `cc-orange-light4B` for this new color in the UI spec so I used that.  If we changed the color names here, then they would no longer sync things up with the UI spec - I think that would be fairly confusing.

emcelroy: OK, that seems very reasonable.

PR #263, opened 2020-12-09
176006585 header nav

mklewandowski:
This PR adds a dropdown menu to the app header that allows the user to navigate between the main view modes (progress view, all responses view, feedback view), updates the all responses view to use the app header, and adds color themes to the app header and its children (so it can be colored to match the view).  

NOTE:  the feedback menu entry will change the header colors, but does not show the feedback panel as that content has not yet been built.

There are a fair amount of changes here so I suggest looking through things in this order:
- add new colors from UI spec (`variables.less`)
- modify SVG icons to handle color changes (several files in the `svg-icons` folder)
- add `viewMode` to the `PortalDashboardApp` (`portal-dashboard-app.tsx`) component state to keep track of which of the 3 views is shown (progress, response, feedback).
- update the `CustomSelect` component (`custom-select.tsx`) to specify an `icon` and an `onSelect` function in the `SelectItems` that are passed in via props.  This is needed for the new navigation dropdown menu layout and to handle changing the view mode.
- update the `Header` component (`header.tsx`) to include a `CustomSelect` component that is used to change the current view mode.
- add support for multiple color options to be used in the `Header` component (`header.tsx`), the `AccountOwner` component (`account-owner.tsx`), the `CustomSelect` component (`custom-select.tsx`), the `HeaderMenu` component (`header-menu.tsx`), and the `HeaderMenuItem` component (`header-menu-item.tsx`)
- remove the `PopupHeader` component and use the main app `Header` component in the `StudentResponsesPopup` component (`student-responses-popup.tsx`).  This required passing in additional props and handling the case where a question or an activity had not yet been specified.
- update the `HeaderMenu` component (`header-menu.tsx`) so clicks outside the menu close the menu.
- remove obsolete `PopupHeader` component
- update Cypress tests

pd-navigation


mklewandowski:
A few other notes on this PR:
- the code climate issues are ones we've seen before caused by duplication (they're pinging us again since I'm working in those areas again).  At some point we might want to try to refactor the upper left nav controls into a single component, but this PR is not the place for that.
- Michael clarified that the upper right menu should not have the compact entry when in the all responses/feedback view.  I'm going to add a new story for that rather than cluttering this already large PR.

emcelroy (approved):
Looks good! Thanks for all the notes about the changes.

PR #264, opened 2020-12-09
Refactor response details and header

mklewandowski:
Refactor the response details and header so that it is easier to work on changes to question list and feedback.


eireland (approved):
Merge it!

PR #265, opened 2020-12-09
Popup question mode

mklewandowski:
Update the popup class nav to support question mode.  Note: this still needs the proper question count and the question sort dropdown to be implemented).

image


eireland (approved):
Merge it!

PR #267, opened 2020-12-10
Change show feeback to hide feedback, fix menu options

mklewandowski:
This PR makes adjustments to the upper right menu:
- change show feedback badges to hide feedback badges (needed to change underlying state as well)
- turn on the feedback badges menu item (will be needed when we implement feedback)
- hid feedback badges and compact menu items if we are on any view other than the progress view


emcelroy (approved):
Looks good to me 👍

PR #269, opened 2020-12-10
Header: Move teacher icon

emcelroy:
PT story: https://www.pivotaltracker.com/story/show/176083674

[#176083674]

* move account icon to the left of teacher name


mklewandowski (approved):
👍

PR #275, opened 2020-12-16
Remove response details animation

mklewandowski:
Remove the animation from the response details view when it is shown to the user. 

Can be demoed here:
http://portal-report.concord.org/branch/response-animation/?portal-dashboard

### DIFF
--- a/css/portal-dashboard/portal-dashboard-app.less
+++ b/css/portal-dashboard/portal-dashboard-app.less
@@ -40,23 +40,4 @@
     flex-direction: column;
   }
 
-  :global .responseDetails-enter {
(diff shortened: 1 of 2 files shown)


emcelroy (approved):
Looks good. I kind of liked that animation, but I guess it's not necessary and is the type of thing that might get tiresome after awhile.

mklewandowski:
> Looks good. I kind of liked that animation, but I guess it's not necessary and is the type of thing that might get tiresome after awhile.

Yeah, I liked it too!  But I think Michael is probably right that it is no longer needed (and time consuming as you point out).

PR #277, opened 2020-12-18
176083480 feedback report view

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/176083480

[#176083480]

In its current state, this draft PR adds a feedback level toggle, feedback badges and feedback badge legend, and feedback textareas to the Feedback Report view. (Right now the textareas aren't functional.)


js/components/portal-dashboard/activity-level-feedback-student-rows.tsx
```diff
@@ -44,7 +46,7 @@ export const ActivityLevelFeedbackStudentRows: React.FC<IProps> = (props) => {
     }
 
     return (
-      <div key={index} className={css.feedbackRows__row}>
+      <div key={activityId + index.toString()} className={css.feedbackRows__row}>
```
kswenson, line 49: As long as the `activityId` is unique, you shouldn't add the `index`. In the case of re-rendering with the order changed or with elements added/removed, the index will change which will cause unnecessary element re-rendering. Using the `activityId` as the key means that the element will re-render iff it is no longer associated with the same activity, which should be what you want.

js/components/portal-dashboard/question-level-feedback-student-rows.tsx
```diff
@@ -48,7 +51,7 @@ export const QuestionLevelFeedbackStudentRows: React.FC<IProps> = (props) => {
     }
 
     return (
-      <div key={index} className={css.feedbackRows__row}>
+      <div key={currentQuestionId + index.toString()} className={css.feedbackRows__row}>
```
kswenson, line 54: Same as above -- as long as the `currentQuestionId` is unique, you shouldn't add the `index`.

js/components/portal-dashboard/activity-level-feedback-student-rows.tsx
emcelroy: @kswenson Thanks. I may be misunderstanding something here, but I think in this case where we have one activity and a row for each student working on that activity, `activityId` won't be unique enough. Each student row should have its own unique key value, right? If we only use `activityId`, wouldn't that key value end up being the same for every row?

js/components/portal-dashboard/question-level-feedback-student-rows.tsx
kswenson: Ah, the importance of understanding the context rather than just looking at a line of code in isolation. In that case, yes, you do need more that the `activityId` or `currentQuestionId`. An ID associated with the student would be better than an index, however, if one is available, for the reasons mentioned above.

js/components/portal-dashboard/activity-level-feedback-student-rows.tsx
kswenson: Ah, the importance of understanding the context rather than just looking at a line of code in isolation. In that case, yes, you do need more that the `activityId` or `currentQuestionId`. An ID associated with the student would be better than an index, however, if one is available, for the reasons mentioned above.

emcelroy: Good point about the student ID option. I'll use that instead. Thanks!

PR #278, opened 2020-12-21
Demo of how to check window redirects

scytacki:
note the code needs to be changed from window.location.href = "something" to window.location.assign("something)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20631,9 +20631,9 @@
       "dev": true
     },
     "query-string": {
-      "version": "6.12.1",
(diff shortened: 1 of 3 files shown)


kswenson (commented):
👍 Looks good -- you should include a link to the blog post in the code so that when it inevitably gets copied elsewhere the link will come along.

scytacki:
@kswenson thanks for the review. I brought this code into the PR that needed it: https://github.com/concord-consortium/portal-report/pull/258/commits/4719d351211decf9c07cddf77af346ee13779d2f

I didn't keep the separate file that was just testing the approach itself.
I did figure out how to throw an exception if `window.location.href=` was called.
I added a link to the blog post.

PR #280, opened 2020-12-23
Logged in user report part 2

scytacki:
Summary of changes:

## If we are redirecting because of OAuth2, do not load the app. 
We are redirecting with `window.location.assign(...)`, this doesn't stop the current javascript from executing. So if we do that redirect then `initializeAuthorization` returns `true`, and the main app is not rendered. See the changes in `js/index.js`.

## Delay the initialization of the firestore database until we know if we are redirecting or not.
Previously the firestore database was initialized immediately in `js/db.ts`. Now it is only initialized if we aren't redirecting. 

## Update access to Firestore db promise.
The previous code would store a promise which resolved to the firestore db and exported `firestoreInitialized`. The other code would chain promises off of that. This way those chained promises would not run until the database was fully initialized. Now `firestoreInitialzed` is replaced with `getFirestore()`. The method returns a promise that resolves to the firestore db like `firestoreInitialized` variable before, but it has the added safety check of throwing an error if that promise hasn't been created yet. The promise is now created by `initializedDB`, and that only happens if we aren't redirecting. (see above)

## expose makeSourceKey
There are a few places in the code where we need to turn URLs into sourceKeys. So this method can now be used instead of adding the custom logic everywhere.

## improve resource url logic
Replace string manipulation with `query-string` module to get the activity parameter. This library automatically decodes url parameters.  This logic was all extracted to its own function to make it easier to understand and should make testing easier.

## delay reading of `firebase-app` url param
Previously the `firebase-app` param was being loaded at the beginning. This meant that it would not be set correctly after we are redirected back to the app and the initial setup of url parameters doesn't include a `firebase-app`. So now there is a `getFirebaseAppName()` function that can be used to look up this value on demand.  Note that the `getPortalFirebaseJWTUrl` function is used for getting JWTs both for the main report as well as JWTs for interactives like Vortex. So it provides a way to override the firebaseApp. Also note that the problem this change is solving wasn't seen yet, because the default firebaseAppName matches what we were setting in the URL parameter.

## use a `portal-report` client id
When OAuth2 is used a client id needs to be sent to the portal, so it can look up the registered redirect urls to see if this is a legitimate request. The code was using a placeholder client id, now it is set to `portal-report`. I looked at the ways this client id was configured in our other SPAs. They all hardcoded the client id into the code, so I'm following that approach here too.

## Code coverage
The remaining uncovered code is around Firestore auth. It isn't easy to test. If we used the firebase emulator we could test it. But it doesn't seem worth introducing that in this PR.

### DIFF
--- a/docs/launch.md
+++ b/docs/launch.md
@@ -41,7 +41,7 @@ In this diagram the Activity Player could be substituted for another third party
 
 - URL: `${auth-domain}/auth/oauth_authorize` This uses the `auth-domain` from the launch params to know where to send the OAuth2 request to.
 - Params:
-  - **clientId** currently hardcoded to `"token-service-example-app"`. This needs to be the name of a client in the portal that has a redirect url which exactly matches the `redirectUri` below. And the client "client type" needs to be configured as 'public', so it allows browser requests for access_tokens.
+  - **clientId** currently hardcoded to `"portal-report"`. This needs to be the `App Id` a client in the portal that has a redirect url which exactly matches the `redirectUri` below. And the client "client type" needs to be configured as 'public', so it allows browser requests for access_tokens.
   - **redirectUri** current portal-report url without any parameters (this way it will include any version or branch path)
   - **state** this is a random string that is used as a key to restore all of the params from the launch after OAuth2 is done. The state has a size limit so the params themselves cannot be sent directly as the state.
 - Code Location: `api.ts#authorizeInPortal`
--- a/docs/launch.md
+++ b/docs/launch.md
@@ -29,7 +29,7 @@ In this diagram the Activity Player could be substituted for another third party
 #### Initial Launch Params (#1 in the diagram)
 
 - **class** URL to portal class api to get info about the class
-- **firebase-app** firebase app name in the portal. It defaults to "report-service". You might set this to work ith the development firestore database with "report-service-dev"
+- **firebase-app** firebase app name in the portal. It defaults to "report-service". You might set this to work with the development firestore database with "report-service-dev"
 - **offering** URL used to request info from the portal about this assignment in the portal
 - **reportType** always set to `offering`
 - **studentId** user id of the student this report is being opened for
@@ -41,14 +41,14 @@ In this diagram the Activity Player could be substituted for another third party
 
 - URL: `${auth-domain}/auth/oauth_authorize` This uses the `auth-domain` from the launch params to know where to send the OAuth2 request to.
 - Params:
-  - **clientId** currently hardcoded to `"token-service-example-app"`. This needs to be the name of a client in the portal that has a redirect url which exactly matches the redirectUri below. And the client "client type" needs to be configured as 'public', so it allows browser requests for access_tokens.
+  - **clientId** currently hardcoded to `"token-service-example-app"`. This needs to be the name of a client in the portal that has a redirect url which exactly matches the `redirectUri` below. And the client "client type" needs to be configured as 'public', so it allows browser requests for access_tokens.
   - **redirectUri** current portal-report url without any parameters (this way it will include any version or branch path)
   - **state** this is a random string that is used a key to restore all of the params from the launch after OAuth2 is done. The state has a size limit so the params be sent directly as the state.
 - Code Location: `api.ts#authorizeInPortal`
 
 #### Source Keys in Firestore
 
-This is brief description of source keys used by the portal report. A more detailed description is in [source-key.md](source-key.md)
+This is a brief description of source keys used by the portal report. A more detailed description is in [source-key.md](source-key.md)
 
 
 ##### Resource Structure Source Key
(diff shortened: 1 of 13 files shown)


docs/source-key.md
```diff
@@ -0,0 +1,127 @@
...
+
+All data would have to change. AP, portal-report, LARA, and researcher report would need to support this new approach, and ideally also be backward compatible with old approach so we can do the migration easily.
+
+### What about the access rules?
+The source is not currently checked by any of the access rules.
+
+### What about conflicting documents?  
+If we don't use the source then the activity-player and LARA answer documents will be mixed in the same folder.
+
+The answer documents currently include a `tool_id`. In the AP this domain plus the path of the url of the running the activity player so it also includes the version or branch in. LARA always adds this to the answers, and it typically should be `https://authoring.concord.org` or a developer might set it to something else for testing purposes.
```
kswenson, line 104: I didn't follow the second sentence: `In the AP this domain plus the path of the url of the running the activity player so it also includes the version or branch in.`

docs/source-key.md
```diff
@@ -0,0 +1,127 @@
...
+
+So this is a way to differentiate between the two types of answer documents.
+
+The question_id in the answer documents is not globably unique by itself, but combining it with the resource_url should uniquely identify the question from the authored content. The same question might be shared by some LARA and AP answers, so a search of answers using just `question_id` and `resource_url` could turn up some answer documents from both systems. However, any reporting system should also be using a `resource_link_id` and/or `run_key` these fields should constrain the answers to a single system.  There are cases that come up with migration where some answers from both systems might share a `resource_link_id`, but that is intentional in the case of a migration.
```
kswenson, line 108:
```suggestion
The question_id in the answer documents is not globally unique by itself, but combining it with the resource_url should uniquely identify the question from the authored content. The same question might be shared by some LARA and AP answers, so a search of answers using just `question_id` and `resource_url` could turn up some answer documents from both systems. However, any reporting system should also be using a `resource_link_id` and/or `run_key` these fields should constrain the answers to a single system.  There are cases that come up with migration where some answers from both systems might share a `resource_link_id`, but that is intentional in the case of a migration.
```

docs/source-key.md
```diff
@@ -0,0 +1,127 @@
...
+
+Perhaps a researcher wants to look at answers to a question across many classes. In that case the query would be:
+`resource_url = https://example.com and question_id = 234`
+That could then pick up answers from different systems. That is hopefully what the researcher would want beause it indicates the same question was given in multiple systems. If a researcher wants to limit the search to just answers created by a particular runtime they'd need to include the `tool_id` in the query.
```
kswenson, line 112:
```suggestion
That could then pick up answers from different systems. That is hopefully what the researcher would want because it indicates the same question was given in multiple systems. If a researcher wants to limit the search to just answers created by a particular runtime they'd need to include the `tool_id` in the query.
```

docs/source-key.md
scytacki: I updated this section to hopefully make it more clear.

docs/launch.md
```diff
@@ -26,70 +26,82 @@ The portal report can run in multiple different modes depending on which URL par
...
+- **answersSourceKey** the source key for the answers. See URL Parameters section below for more info.
+- **auth-domain** URL for the authentication domain, usually the portal so it would usually be `https://learn.concord.org`
+
+#### Request to portal OAuth2 (#3 in the diagram)
+
+- URL: `${auth-domain}/auth/oauth_authorize` This uses the `auth-domain` from the launch params to know where to send the OAuth2 request to.
+- Params:
+  - **clientId** currently hardcoded to `"token-service-example-app"`. This needs to be the name of a client in the portal that has a redirect url which exactly matches the redirectUri below. And the client "client type" needs to be configured as 'public', so it allows browser requests for access_tokens.
+  - **redirectUri** current portal-report url without any parameters (this way it will include any version or branch path)
+  - **state** this is a random string that is used a key to restore all of the params from the launch after OAuth2 is done. The state has a size limit so the params be sent directly as the state.
```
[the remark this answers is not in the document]
scytacki: I updated this in a separate commit.

scytacki:
@eireland this is ready to be reviewed now.

scytacki:
@mklewandowski I added you as a reviewer because you looked at the activity player side of these changes. But if you don't want to review it, that is also fine. @eireland and I have been working on it together, so it has had at least 2 pairs of eyes.

mklewandowski:
> @mklewandowski I added you as a reviewer because you looked at the activity player side of these changes. But if you don't want to review it, that is also fine. @eireland and I have been working on it together, so it has had at least 2 pairs of eyes.

I'll take a look at it today.  Thanks!

mklewandowski:
One thing that appears to now be broken is the handling of activity level feedback.  This feature was broken BEFORE this PR (I'm guessing in one of the rounds of recent work), but might be related to some of the work in this PR.  When we try to access activity level feedback in the old report, we get  `Uncaught APIError {message: "No token available", response: {…}}`.  
image

@scytacki does it seem like this error is tied to this work?

docs/source-key.md
```diff
@@ -0,0 +1,127 @@
+# Source Key
+
+In the report service database the data is segmented in to 'sources'. Each source has a 'sourceKey'.
+
+They were original intended to segment the data stored in the report-service. This way tools working with the report-service wouldn't accidentally clobber some other tool's data. However, we are now sharing this data between multiple tools, so they aren't as useful as they used to be.
```
mklewandowski, line 5: Minor typo fix:
`They were originally intended to segment the data stored in the report-service.`

eireland (approved):

scytacki:
@mklewandowski thanks for pointing that out. I'll go ahead and merge this since the same issue is happening on master. I'll make a new story and look at the issue more.

PR #281, opened 2020-12-24
176081835 student link

eireland:
Adds ability open the response view when user clicks on student name. Response View opens in question mode of the student clicked on

### DIFF
--- a/js/components/portal-dashboard/question-overlay.tsx
+++ b/js/components/portal-dashboard/question-overlay.tsx
@@ -14,6 +14,7 @@ interface IProps {
   currentQuestion?: Map<string, any>;
   currentStudentId: string | null;
   setDashboardViewMode: (mode: DashboardViewMode) => void;
+  setListViewMode: (mode: boolean) => void;
   isAnonymous: boolean;
   questions?: Map<string, any>;
   setCurrentActivity: (activityId: string) => void;
@@ -45,7 +46,7 @@ export class QuestionOverlay extends React.PureComponent<IProps> {
   }
 
   private renderQuestionDetails = () => {
-    const { currentActivity, currentQuestion, questions, sortedQuestionIds, toggleCurrentQuestion, setCurrentActivity, hasTeacherEdition } = this.props;
+    const { currentQuestion, questions, sortedQuestionIds, toggleCurrentQuestion, setCurrentActivity, hasTeacherEdition } = this.props;
     return (
       <React.Fragment>
         <div className={css.header} data-cy="question-overlay-header">
@@ -56,7 +57,6 @@ export class QuestionOverlay extends React.PureComponent<IProps> {
           {this.renderAllResponsesButton()}
         </div>
         <QuestionNavigator
-          currentActivity={currentActivity}
           currentQuestion={currentQuestion}
           questions={questions}
           sortedQuestionIds={sortedQuestionIds}
@@ -86,5 +86,6 @@ export class QuestionOverlay extends React.PureComponent<IProps> {
 
   private handleShowAllResponsesButtonClick = () => {
     this.props.setDashboardViewMode("ResponseDetails");
+    this.props.setListViewMode(false);
   }
 }
--- a/js/components/portal-dashboard/question-overlay.tsx
+++ b/js/components/portal-dashboard/question-overlay.tsx
@@ -5,7 +5,7 @@ import { QuestionNavigator } from "./question-navigator";
 import { StudentResponse } from "./overlay-student-response";
 import GroupIcon from "../../../img/svg-icons/group-icon.svg";
 import QuestionPopoutIcon from "../../../img/svg-icons/question-popout-icon.svg";
-import { DashboardViewMode } from "../../util/misc";
+import { DashboardViewMode, ListViewMode } from "../../util/misc";
 
 import css from "../../../css/portal-dashboard/question-overlay.less";
 
@@ -14,7 +14,7 @@ interface IProps {
   currentQuestion?: Map<string, any>;
   currentStudentId: string | null;
   setDashboardViewMode: (mode: DashboardViewMode) => void;
-  setListViewMode: (mode: boolean) => void;
+  setListViewMode: (mode: ListViewMode) => void;
   isAnonymous: boolean;
   questions?: Map<string, any>;
   setCurrentActivity: (activityId: string) => void;
@@ -86,6 +86,6 @@ export class QuestionOverlay extends React.PureComponent<IProps> {
 
   private handleShowAllResponsesButtonClick = () => {
     this.props.setDashboardViewMode("ResponseDetails");
-    this.props.setListViewMode(false);
+    this.props.setListViewMode("Student");
   }
 }
--- a/js/components/portal-dashboard/response-details/response-details.tsx
+++ b/js/components/portal-dashboard/response-details/response-details.tsx
@@ -8,6 +8,7 @@ import { SpotlightStudentListDialog, spotlightColors } from "./spotlight-student
 import { TransitionGroup, CSSTransition } from "react-transition-group";
 import { StudentNavigator } from "../student-navigator";
 import { ActivityNavigator } from "../activity-navigator";
+import { ListViewMode } from "../../../util/misc";
 
 import css from "../../../../css/portal-dashboard/response-details/response-details.less";
 import { PopupQuestionAnswerList } from "./popup-question-answer-list";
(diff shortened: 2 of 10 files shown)


js/components/portal-dashboard/question-overlay.tsx
```diff
@@ -14,6 +14,7 @@ interface IProps {
   currentQuestion?: Map<string, any>;
   currentStudentId: string | null;
   setDashboardViewMode: (mode: DashboardViewMode) => void;
+  setListViewMode: (mode: boolean) => void;
```
mklewandowski, line 17: This name could be improved.  `setListViewMode` implies that it will take a `mode`, but it takes a `boolean`.

js/containers/portal-dashboard/portal-dashboard-app.tsx
```diff
@@ -229,6 +237,10 @@ class PortalDashboardApp extends React.PureComponent<IProps, IState> {
     this.setState({ viewMode: mode });
   }
 
+  private setListViewMode = (value: boolean) => {
```
mklewandowski, line 240: Here's the origin of the function.  This should probably be something like `toggleQuestionMode`.  Since you're not setting a mode, you're toggling a boolean.  I would mass change the instances of `setListViewMode` to something more in line with what the function does (like `toggleQuestionMode`).

js/components/portal-dashboard/student-navigator.tsx
```diff
@@ -20,8 +20,8 @@ export const StudentNavigator: React.FC<IProps> = (props) => {
                         ? [<StudentName key={"student-name"} {...props}/>, <PrevNextButtons key={"prev-next-button"} {...props}/>]
                         : [<PrevNextButtons key={"prev-next-button"} {...props}/>, <StudentName key={"student-name"} {...props}/>];
   return (
-    <div className={css.studentNavigator}>
-      <div className={css.header}>
+    <div className={css.studentNavigator} >
```
mklewandowski, line 23: probably don't want that added space to be consistent with rest of app.

js/components/portal-dashboard/student-names.tsx
```diff
@@ -21,12 +25,18 @@ export class StudentNames extends React.PureComponent<IProps> {
           return (
             <div className={`${css.studentName} ${compactClass}`} key={`student ${i}`}>
               <div key={`student ${i}`}>
-                <div className={css.name} data-cy="student-name">{formattedName}</div>
+                <div className={css.name} data-cy="student-name" onClick={this.handleOpenStudentView("ResponseDetails", student.get("id"))}>{formattedName}</div>
```
mklewandowski, line 28: Line length is getting pretty long.  Might want something like:
```
                <div 
                  className={css.name} 
                  data-cy="student-name" 
                  onClick={this.handleOpenStudentView("ResponseDetails", student.get("id"))}
                >
                    {formattedName}
                </div>
```

mklewandowski (changes requested):
Looks good!  I had a couple of minor suggestions that shouldn't take long.  I think we want to at least rename that function so it has a clearer name (along with the prop names in components where it is passed).

js/components/portal-dashboard/question-overlay.tsx
mklewandowski: See my comments below about naming this `toggleQuestionMode` - unless this somehow clashes with upcoming feedback work.

js/components/portal-dashboard/response-details/response-details.tsx
```diff
@@ -25,12 +25,14 @@ interface IProps {
...
   currentStudentId: string | null;
   hasTeacherEdition: boolean;
+  inQuestionMode: boolean;
   isAnonymous: boolean;
   questions?: Map<string, any>;
   setAnonymous: (value: boolean) => void;
   setCurrentActivity: (activityId: string) => void;
   setCurrentQuestion: (questionId: string) => void;
   setCurrentStudent: (studentId: string | null) => void;
+  setListViewMode: (mode: boolean) => void;
```
mklewandowski, line 35: See my comments below about naming this `toggleQuestionMode` - unless this somehow clashes with upcoming feedback work.

js/components/portal-dashboard/student-names.tsx
```diff
@@ -1,12 +1,16 @@
...
 
 import css from "../../../css/portal-dashboard/student-names.less";
 
 interface IProps {
   students: any; // TODO: add type
   isAnonymous: boolean;
   isCompact: boolean;
+  setCurrentStudent: (studentId: string) => void;
+  setDashboardViewMode: (mode: DashboardViewMode) => void;
+  setListViewMode: (mode: boolean) => void;
```
mklewandowski, line 13: See my comments below about naming this `toggleQuestionMode` - unless this somehow clashes with upcoming feedback work.

mklewandowski:
One note, I would need to know more about the upcoming work, but is there some chance that the `setListViewMode` function will eventually be setting one of three states once we add feedback?  If so, then we may consider leaving the name.  However, if it is just toggling the question boolean, then I would rename it `toggleQuestionMode`.

js/components/portal-dashboard/student-names.tsx
eireland: How about `showQuestionMode`?

mklewandowski: It's more like a set or toggle though, isn't it?  We're not showing/hiding question mode.  We're either in question mode or we're in student mode, but in this case we're trying to handle it with a single boolean.  Show/hide sort of implies that a piece of the UI is either there or isn't there.  
Ultimately this probably would be best if it remained `setListViewMode`, but rather than taking a boolean, it took a `ListViewMode` type.  You could model it exactly as I implemented `DashboardViewMode`.  So in misc.ts, add something like:
`export type ListViewMode = "Question" | "Student";`

In app.tsx instead of `inQuestionMode: boolean;` you would have `listViewMode : ListViewMode`.  And then setListViewMode would be something like:
```
  private setListViewMode = (value: ListViewMode) => {
    this.setState({ listViewMode : value });
  }
```

That is really the cleanest way to do it.  Looking now, that doesn't seem to be much more work than changing the name of the function.  So I would to leave `setListViewMode`, but change it so it takes a `ListViewMode` type.  And then have the list view mode stored in state in app.tsx as `listViewMode: ListViewMode`;

mklewandowski: This would allow us to expand the list view mode in the future if needed and it more clearly describes what is happening.  We have to list view mode types and we are setting one of the other when a button is pressed.  The UI is then updated based on the list view mode type.  Much cleaner, more descriptive, and extensible than using a boolean.

js/components/portal-dashboard/response-details/response-details.tsx
mklewandowski: Per my comment below, I think the best approach is to change this to:
`setListViewMode: (mode: ListViewMode) => void;`

js/containers/portal-dashboard/portal-dashboard-app.tsx
mklewandowski: Per my comment below, I think the best approach is to change this to:
`private setListViewMode = (value: ListViewMode) => {`

js/components/portal-dashboard/question-overlay.tsx
mklewandowski: Per my comment below, I think the best approach is to change this to:
`setListViewMode: (mode: ListViewMode) => void;`

js/components/portal-dashboard/student-names.tsx
```diff
@@ -21,12 +25,22 @@ export class StudentNames extends React.PureComponent<IProps> {
...
         }) }
       </div>
     );
   }
+
+  private handleOpenStudentView = (mode: DashboardViewMode, currentStudentId: string) => () => {
+    this.props.setDashboardViewMode(mode);
+    this.props.setListViewMode("Question");
+    this.props.setCurrentStudent(currentStudentId);
+  }
```
mklewandowski, line 45: Is this function named correctly?  We call handleOpenStudentView and set the list view mode to question.  That seems backwards.

mklewandowski: Ah, I see the issue, we click on a student name and open the question view mode for that student.  I think this is a remnant of the previous naming.  We're really opening responseDetails in Question mode.  I would consider calling this something like handleOpenResponseDetailsInQuestionMode.  Or something that is more clear about what it does.

mklewandowski (approved):
👍 Looks good.  I would consider renaming handleOpenStudentView since we're actually opening the Response Details view in Question mode.  The function doesn't really do what the name implies.

js/components/portal-dashboard/student-names.tsx
eireland: It's that way because ListViewMode "Student" lists the studens and their answer to the one question. Versus ListViewMode "Question" lists the questions and their answers from the activitiy for the one particular student. The icons also show the same thing, person outline for the "Student", "Q" for the "Question"

PR #282, opened 2020-12-31
Adds ability to show fake activity data with query param

eireland:
User can now add `resourceType=activity` query param to show activity resource and answer. It defaults to showing sequence data if query param is not present.
This should make it easier to switch between fake sequence and fake activity data without having to change it in code.

### DIFF
--- a/js/actions/index.ts
+++ b/js/actions/index.ts
@@ -1,8 +1,8 @@
 import { firestoreInitialized } from "../db";
 import fakeSequenceStructure from "../data/sequence-structure.json";
-// import fakeActivityStructure from "../data/activity-structure.json";
+import fakeActivityStructure from "../data/activity-structure.json";
 import fakeAnswers from "../data/answers.json";
-// import fakeAnswers from "../data/average-class-activity-answers.json";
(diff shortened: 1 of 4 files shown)


js/actions/index.ts
```diff
@@ -122,19 +122,27 @@ function _receivePortalData(db: firebase.firestore.Firestore,
...
-    //   type: RECEIVE_RESOURCE_STRUCTURE,
-    //   response: fakeActivityStructure,
-    // });
-    dispatch({
-      type: RECEIVE_ANSWERS,
-      response: fakeAnswers,
-    });
+    // Use fake data. Default shows sequence fake resource and answer
+    // resourceType query param allows for switching to show fake activity resource and answer
+    if(urlStringParam(window.location.search, "resourceType")==="activity") {
```
mklewandowski, line 127: spacing:
```
if (urlStringParam(window.location.search, "resourceType") === "activity") {
```

mklewandowski (approved):
👍 LGTM!  This is a nice improvement that will make things much easier to work with.  I had one suggestion about spacing on a line in index.ts, but that should be a simple fix.

PR #284, opened 2021-01-06
176329010 feedback badge legend

eireland:
Adds the feedback badges to the legend container in the progress view. When you click on Hide Feedback Badges in the header menu, the feedback legend should get grayed out.


js/components/portal-dashboard/legend-container.tsx
```diff
@@ -1,12 +1,12 @@
 import React from "react";
 
-import css from "../../../css/portal-dashboard/progress-legend.less";
+import css from "../../../css/portal-dashboard/progress-view-legend.less";
 
-export interface ProgressType {
-  name: string;
+export interface LegendType {
+  name: string | null;
```
mklewandowski, line 6: You could probably also write this simply as:
`name?: string;`

js/components/portal-dashboard/legend-container.tsx
```diff
@@ -22,20 +22,61 @@ export const progress: ProgressType[] =
     }
   ];
 
-export class ProgressLegendContainer extends React.PureComponent {
+  export const feedback: LegendType[] =
+  [
+    {
+      name: null,
```
mklewandowski, line 28: And if you write `name?: string;` above, you can remove this.

js/components/portal-dashboard/legend-container.tsx
```diff
@@ -22,20 +22,61 @@ export const progress: ProgressType[] =
...
+                <div className={css.legendText}>{progress.name}</div>
+              </div>
+            );
+          })}
+        </div>
+        <div className={css.divider}></div>
+        <div className={css.feedbackLegend} data-cy="feedback-legend">
+          <div className={css.legendLabel}>Feedback:</div>
+          { feedback.map((feedback, index) => {
+            const feedbackName = feedback.name? feedback.name.replace(/\ /g,'-') : "no-title";
```
mklewandowski, line 64: spacing:
`const feedbackName = feedback.name ? feedback.name.replace(/\ /g,'-') : "no-title";`

js/components/portal-dashboard/legend-container.tsx
```diff
@@ -22,20 +22,61 @@ export const progress: ProgressType[] =
...
+            return (
+              <div key={index}
+                   className={`${css.legendKey} ${this.props.hideFeedbackBadges? css.disabled: ""}`}
+                   data-cy={`${feedbackName}-legend${this.props.hideFeedbackBadges ? "-disabled" : ""}`}>
+                  {feedback.name === "Answer Updated"
+                    ? <div className={`${css.legendIcon} ${css.feedbackIcon} ${css.questionFeedbackGiven}`}>
+                        <div className={`${feedback.class}`}></div>
```
mklewandowski, line 71: Don't need string template:
```
<div className={feedback.class}></div>
```

js/components/portal-dashboard/legend-container.tsx
```diff
@@ -22,20 +22,61 @@ export const progress: ProgressType[] =
...
+        <div className={css.progressLegend} data-cy="progress-legend">
+          <div className={css.legendLabel}>Activity Progress:</div>
+          { progress.map((progress, index) => {
+            const progressName = progress.name? progress.name.replace(/\ /g,'-') : "no-title";
+            return (
+              <div key={index} className={css.legendKey} data-cy={progressName + "-legend"}>
+                <div className={`${css.legendIcon} ${css.progressIcon} ${progress.class}`}></div>
+                <div className={css.legendText}>{progress.name}</div>
+              </div>
+            );
```
mklewandowski, line 57: This chunk might be a little easier to read with a slight tweak in spacing:
```
            return (
              <div 
                key={index}
                className={`${css.legendKey} ${this.props.hideFeedbackBadges? css.disabled: ""}`}
                data-cy={`${feedbackName}-legend${this.props.hideFeedbackBadges ? "-disabled" : ""}`}
              >
                { feedback.name === "Answer Updated"
                  ? <div className={`${css.legendIcon} ${css.feedbackIcon} ${css.questionFeedbackGiven}`}>
                      <div className={feedback.class}></div>
                    </div>
                  : <div className={`${css.legendIcon} ${css.feedbackIcon} ${feedback.class}`}></div>
                }
                <div className={`${css.legendText} ${!feedback.name && css.noText}`}>{feedback.name}</div>
              </div>
            );
```

mklewandowski (approved):
👍 Looks good!  I had a couple of small suggestions that should be easy to slot in.  One question, is the disabled state (when badges are hidden) in the spec somewhere?  I couldn't find it.

js/components/portal-dashboard/legend-container.tsx
kswenson: While that suggestion may be correct, it's not equivalent given that `null` !== `undefined`.

mklewandowski: Agreed, not equivalent, but I'm suggesting the removal of null entirely in the comment below.  We will accomplish the same thing since this code is just trying to determine if there is a name or not (the use of null doesn't do anything unique from using undefined in this case).

eireland:
Spec for disabled badges was in zeplin
image

PR #286, opened 2021-01-08
fix several issues with the api error handling

scytacki:
Runtime changes:
- callAPI now handles errors thrown during promise evaluation or when an api function throws an error before returning the promise.
- APIError now extends Error which provides stack traces at the location the error was generated. Extending Error can be problematic because it breaks instanceof checks. That happens due to transpiling. Several articles on this, here is one example: https://www.bennadel.com/blog/3226-experimenting-with-error-sub-classing-using-es5-and-typescript-2-1-5.htm
- to work around the instanceof check above, now the name property is used instead of instanceof to identify APIError objects.
- remove the re-throwing of the error since we are already printing it to the console. There might be edge
cases where some un handled error happens during the callAPI setup, which could stop the code if we threw it again. And maybe that would be good thing in some cases, but I don't think those edge cases are worth the confusion. In the general case the error will happen in a promise so there isn't a way to stop the code.
- update some api functions to be properly async so their errors can be caught using the standard promise.catch
approach.
- fix the DataFetchError component so it actually displays the content of error responses which have bodies.

Test changes:
- Update to Cypress 6.2
- Update cypress tests to support 6.2:
  - `should("not.be.visible")` fails in cypress 6 if the element doesn't exist
  - `cy.get("first").within(() => { cy.get("child of first") })` now yields `first` instead of `child of first`. This matches what the documentation has been saving for a long time: https://docs.cypress.io/api/commands/within.html#Yields but until recently within did not behave this way.  This was "fixed" in Cypress 5.4:  https://docs.cypress.io/guides/references/changelog.html#5-4-0:~:text=When%20a%20command%20is%20chained%20after%20.within()
Our code was assuming that within would return the last subject of the function passed to within. So those cases were changed to using a 'find' call instead.
- added tests for the different ways errors can be thrown in callApi
- added tests of the DataFetchError component
- added tests of the updateReportSettingsInPortal, this is the code path that was causing the issue which is why I got started on this error improvement PR.

### DIFF
--- a/js/components/report/data-fetch-error.js
+++ b/js/components/report/data-fetch-error.js
@@ -8,14 +8,14 @@ export default class DataFetchError extends PureComponent {
     this.state = {};
   }
 
-  // Inorder to reuse updateBodyText we can't call it in the constructor because
+  // In order to reuse updateBodyText we can't call it in the constructor because
   // we can't call setState until the component has been mounted
   componentDidMount() {
     this.updateBodyText();
   }
 
   componentDidUpdate(prevProps) {
-    if(prevProps.error !== this.props.error) {
+    if (prevProps.error !== this.props.error) {
       this.updateBodyText();
     }
   }
--- a/js/components/report/data-fetch-error.js
+++ b/js/components/report/data-fetch-error.js
@@ -3,6 +3,43 @@ import React, { PureComponent } from "react";
 import "../../../css/report/data-fetch-error.less";
 
 export default class DataFetchError extends PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
+  // Inorder to reuse updateBodyText we can't call it in the constructor because
+  // we can't call setState until the component has been mounted
+  componentDidMount() {
+    this.updateBodyText();
+  }
+
+  componentDidUpdate(prevProps) {
+    if(prevProps.error !== this.props.error) {
+      this.updateBodyText();
+    }
+  }
+
+  updateBodyText() {
+    const { error } = this.props;
+    const body = error.body;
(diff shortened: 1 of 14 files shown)


js/components/report/data-fetch-error.js
```diff
@@ -3,6 +3,43 @@ import React, { PureComponent } from "react";
 import "../../../css/report/data-fetch-error.less";
 
 export default class DataFetchError extends PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
+  // Inorder to reuse updateBodyText we can't call it in the constructor because
```
mklewandowski, line 11: "In order"

js/components/report/data-fetch-error.js
```diff
@@ -3,6 +3,43 @@ import React, { PureComponent } from "react";
...
+  // we can't call setState until the component has been mounted
+  componentDidMount() {
+    this.updateBodyText();
+  }
+
+  componentDidUpdate(prevProps) {
+    if(prevProps.error !== this.props.error) {
```
mklewandowski, line 18: Minor spacing:
` if (prevProps.error !== this.props.error) {`

mklewandowski (approved):
👍 I did a cursory review and it looks good (I added a couple of comments about minor spacing), but I'm not entirely familiar with this code.  It's probably worth having another reviewer take a look.

scytacki:
Thanks @mklewandowski I don't think anyone is familiar with this code. :)

As a tip when you make single line corrections you can use the 
image button. And then I can apply your corrections right in the github UI, and it also adds you to git history as the contributor of the commit.

PR #287, opened 2021-01-08
fix local feedback testing

scytacki:
this was causing problems when testing because we had no token in the url
so getAuthHeader was throwing an error

This can be tested with:
http://portal-report.concord.org/branch/fix-error-with-feedback/?enableFirestorePersistence=true

To track down this really simple change, I improved the error handling, and I put that in a different PR:
https://github.com/concord-consortium/portal-report/pull/286

### DIFF
--- a/test/api_spec.js
+++ b/test/api_spec.js
@@ -1,6 +1,6 @@
 import nock from "nock";
 import { fetchOfferingData, getPortalFirebaseJWTUrl, fetchFirestoreJWT,
-  initializeAuthorization, getAuthHeader } from "../js/api";
+  initializeAuthorization, getAuthHeader, updateReportSettingsInPortal } from "../js/api";
 import queryString from "query-string";
 
 describe("api helper", () => {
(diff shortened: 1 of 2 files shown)


scytacki:
I looked into why codecov is reporting reduced coverage. 
My modified code is fully covered. 
The negative changes are in other parts of the code:
- `js/components/dashboard/dashboard.js` did not cover the scrolling functions:
  - `onHorizontalContainerScroll`
  - `onHeadersScroll`
  - `ignoreNextScrollEvent`
  I suspect this is because it is random whether the cypress tests trigger a scroll or not
- `js/api-middleware.js` is not covering a the `throw error` now. This is expected and that line is removed in the related PR.

The solution for the scrolling issue should be that we need an explicit scrolling test in cypress, but I think that ought to be a in a different PR.

test/api_spec.js
```diff
@@ -77,6 +77,65 @@ describe("api helper", () => {
...
+  describe("updateReportSettingsInPortal", () => {
+    // need a fake offering url param that includes /offerings/
+    const fakeOfferingUrl = "https://portal.com/offerings/123";
+    const okResponse = "OK";
+    const fakeSettings = {something: "important"};
+
+    beforeEach(() => {
+    });
+
+    describe("when offering and token URL param is present", () => {
```
emcelroy, line 89: Would it be a little bit more clear to say something like "when both offering and token URL params are present"?

emcelroy (approved):
Looks good 👍 I had one comment/question, but I don't think it's vitally important.

test/api_spec.js
scytacki: Yup, thanks.

PR #292, opened 2021-01-21
Feedback note popup

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/176161785

[#176161785]

* Feedback note button has default, hover, and selected states
* When button is clicked, open popup dialog
* Popup has a button to close the popup that reads "Got it" with default, hover, and pressed states

### DIFF
--- a/css/portal-dashboard/feedback/feedback-note-modal.less
+++ b/css/portal-dashboard/feedback/feedback-note-modal.less
@@ -1,0 +1,77 @@
+@import "../variables";
+
+.lightbox {
+  background-color: white;
+  border-radius: 8px !important;
+  box-shadow: 0 0 5px 0 rgba(0, 0, 0, 0.35);
+  height: auto;
+  left: 160px;
+  top: 135px;
+  width: 320px;
+
+  .lightboxHeader {
+    border-radius: 8px 8px 0 0;
+    overflow: auto;
+    padding: 0;
+
+    .title {
+      color: @feedback-green;
+      font-weight: bold;
+      height: 24px;
+      line-height: 1.4;
+      margin: 20px 20px 15px;
+      padding-left: 29px;
+      position: relative;
+      text-align: left;
+
+      &:before {
+        background-color: @feedback-green;
+        content: "";
+        display: block;
+        height: 24px;
+        left: 0;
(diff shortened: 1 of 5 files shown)


mklewandowski (approved):
👍 This looks great to me!  The only thing I wonder is should the note button itself have an active state when pressed?  I looked in the UI spec and didn't find anything, but knowing Michael, it would probably use the same colors that are used when the modal is showing (green background, white text).

mklewandowski:
Looks like the prototype does have an active style of the note button:
image

emcelroy:
Thanks, @mklewandowski. There is an `active` class for that button in the code. You can see the style rules in css/portal-dashboard/feedback/feedback-note-toggle.less starting on line 20. The class is applied if the boolean `buttonActive` is true in js/components/portal-dashboard/feedback/feedback-note-toggle.tsx (lines 15, 18, 23). It works (see GIF), but is how I did that maybe not as straightforward as it could be?

active-state

Also, I switched this PR back to a draft because of a coverage problem that showed up after I merged in master. I'm having a little trouble tracking that down. Hopefully I can get that figured out today.

mklewandowski:
> Thanks, @mklewandowski. There is an `active` class for that button in the code. You can see the style rules in css/portal-dashboard/feedback/feedback-note-toggle.less starting on line 20. The class is applied if the boolean `buttonActive` is true in js/components/portal-dashboard/feedback/feedback-note-toggle.tsx (lines 15, 18, 23). It works (see GIF), but is how I did that maybe not as straightforward as it could be?
> 
> 
> 
> Also, I switched this PR back to a draft because of a coverage problem that showed up after I merged in master. I'm having a little trouble tracking that down. Hopefully I can get that figured out today.

Right, I see it working when the popup is showing.  That looks good.  But based on the prototype, it also looks like the state should change when the mouse is down over the button before the popup has shown (which is typical with Michael's specs - a mouse hover and mouse down state).  We just need to add some CSS that uses the :active pseudo class to get this behavior.

emcelroy:
Oh, I see. Thanks for catching that missing nuance. I've added it with the latest commit.

mklewandowski (approved):
👍 Looks good!  Thanks for the fix!

PR #293, opened 2021-01-21
176162207 feedback autosave

mklewandowski:
This PR adds a 2 second throttle on the text input of the question and activity level feedback (only save feedback every 2 seconds).


js/components/portal-dashboard/feedback/activity-feedback-textarea.tsx
```diff
@@ -20,20 +21,29 @@ export const ActivityFeedbackTextarea: React.FC<IProps> = (props) => {
...
-      updateActivityFeedback(activityId, activityIndex, studentId, {feedback: target.value});
-      setHeight(target.scrollHeight);
+      updateActivityFeedback(activityId, activityIndex, studentId, {feedback: textareaRef.current?.value});
+      // eslint-disable-next-line no-console
+      console.log(textareaRef.current?.value);
     }
   };
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const updateFeedbackThrottled = useCallback(throttle(updateFeedback, 2000), []);
```
kswenson, line 39: I would think we would want debounce here rather than throttle. Throttle will save immediately on first change and then wait two seconds while debounce will wait until two seconds have elapsed before saving.

mklewandowski: I thought about using debounce at first too, but the story clearly states to save the user input data every 2 seconds.  With debounce, a user can continuously type for 60 seconds and none of their feedback is saved until after the typing is completed (so if they lost internet or their browser crashed at second 59, nothing is saved).  With throttle, we get the save every two seconds.  I also added an onBlur to deal with the final save if the user quickly moves off of the text area and navigates elsewhere.

emcelroy (approved):
This looks good to me! I am curious to see how the debounce vs. throttle conversation goes, though.

As I mentioned elsewhere, it seemed to work great when I tested it locally. The only time it wouldn't save everything was if in the middle of typing I did a quick Command+R to reload the page. That doesn't seem worth trying to protect against though.

PR #294, opened 2021-01-22
Add feedback review flags

mklewandowski:
This PR adds the review flag when submitting feedback which allows us to determine if feedback has been submitted and it it needs to be updated by examining the feedback in the Redux state.  Although feedback was being submitted to the student, we we're marking these internal feedback flags.


emcelroy (approved):
Looks good -- thanks!

eireland (approved):
merge it!

PR #297, opened 2021-01-24
Fix issues with activity navigator styling.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/176620502

[#176620502]

These changes fix a couple small issues within the activity navigator section. They change the background color from orange to teal when working with activity-level feedback, and ensure the activity title and nav elements don't appear when working with a standalone activity (as opposed to an activity within a sequence).


mklewandowski (approved):
👍 This looks good to me!

PR #298, opened 2021-01-24
Remove manual resize option for input fields.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/176083561

[#176083561]

This removes the manual resize option for the input fields per the PT story note that says if manual resizing adds complexity, only use the dynamic, responsive option.


mklewandowski (approved):
👍

PR #299, opened 2021-01-25
Added logging of dashboard usage (#174653677)

dougmartin:


dougmartin:
@scytacki I added logging of the activity and question level feedback now that it is merged.  I was waiting for that before I had this PR reviewed.  The history is a little off as I had to rebase master to pull in the feedback changes.

scytacki (changes requested):
If you can change the activity property then I think Trudi can test this out.

I don't understand how the sendToLogManager works. Does it use the 'logging=true" url parameter?

dougmartin:
@scytacki I've added a follow on commit to check the logging flag and set the activity name in the log message: https://github.com/concord-consortium/portal-report/pull/299/commits/f0c319caf6d69968421af7e993fbbf958df965eb

scytacki (approved):
Looks good to me.

PR #301, opened 2021-01-26
Activity rubric feedback

eireland:
Adds activity level rubric feedback when rubric feedback is available.
Student rating can be turned selected and deselected.
When a rating is selected, the feedback is submitted, and the feedback icon should show feedback is given. If none of the ratings are selected, the feedback icon should show awaiting feedback.


.eslintrc.js
```diff
@@ -42,7 +42,7 @@ module.exports = {
     "eol-last": "warn",
     eqeqeq: ["error", "smart"],
     "no-case-declarations": "off",
-    "no-console": ["error", { allow: ["warn", "error"] }],
+    // "no-console": ["error", { allow: ["warn", "error"] }],
```
emcelroy, line 45: I think this change must've gotten checked in by accident, right?

css/portal-dashboard/feedback/feedback-rows.less
```diff
@@ -77,8 +77,8 @@
         min-height: 50px;
         outline: none;
         padding: 5px 10px 13px;
-        resize: none;
-        width: calc(100% - 20px);
+        resize: vertical;
```
emcelroy, line 80: `resize` should be set to `none` as of this PR https://github.com/concord-consortium/portal-report/pull/298.

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
+import React from "react";
+// import { Map } from "immutable";
```
emcelroy, line 2: Should this commented out line be removed?

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  private renderColumnHeaders = (rubric: any) => {
+    const { referenceURL } = rubric;
+    return (
+      <div className={css.columnHeaders}>
+        <div className={css.rubricDescriptionHeader}>
+          <div className={css.scoringGuideArea}>
+            <a className={css.launchButton}>
+              <LaunchIcon className={css.icon} href={referenceURL} target="_blank" />
```
emcelroy, line 49: This link/button isn't working. I think the `href` needs to be moved from the icon to the `a` element.

emcelroy (changes requested):
This is looking pretty good, but I think it needs some changes. See inline comments.

Another issue is that the pseudo radio buttons don't seem to work fully. When I tried this out locally and clicked on one, nothing happened and there was an error in the console along the lines of `Uncaught TypeError: Cannot read property 'C1' of undefined at rubric-table.tsx:107`.

.eslintrc.js
mklewandowski: I'll second this.  We want to leave this lint setting on.

css/portal-dashboard/feedback/rubric-table.less
```diff
@@ -0,0 +1,167 @@
...
+            height: 24px;
+            pointer-events: none;
+            fill: @feedback-green;
+          }
+        }
+      }
+
+      .rubricDescriptionTitle {
+        justify-content: center;
+        width: 75%
```
mklewandowski, line 65: Percents like this can result in odd effects at big and small widths.  That said, this looked good in practice.

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
+import React from "react";
+// import { Map } from "immutable";
+import Markdown from "markdown-to-jsx";
+import LaunchIcon from "../../../../img/svg-icons/launch-icon.svg";
+
+import css from "../../../../css/portal-dashboard/feedback/rubric-table.less";
```
mklewandowski, line 6: I would add a line break after this import for clarity.

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+interface IProps {
+  activityStarted: boolean;
+  rubric: any;
+  student: any;
+  rubricFeedback: any;
```
mklewandowski, line 11: Do we know types on any of the above three?

css/portal-dashboard/feedback/rubric-table.less
```diff
@@ -0,0 +1,167 @@
...
+            &:hover .outerCircle {
+              background-color: white;
+
+              .innerCircle {
+                background-color: @feedback-green-light3B;
+              }
+            }
+            &:active .outerCircle {
+              background-color: white;
+              // color: @feedback-green;
```
mklewandowski, line 129: Can this be deleted?

package.json
```diff
@@ -72,6 +72,7 @@
     "@types/jest": "^25.2.2",
     "@types/jsonwebtoken": "^8.5.0",
     "@types/lodash": "^4.14.168",
+    "@types/markdown-to-jsx": "^6.11.3",
```
mklewandowski, line 75: Is this used?  Not seeing why this is added.

mklewandowski: Ah, I see it in the rubric table now.

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+  activityId: string;
+  activityIndex: number;
+  updateActivityFeedback?: (activityId: string, activityIndex: number, platformStudentId: string, feedback: any) => void;
```
mklewandowski, line 14: Does this need to be optional? Don't we always need to pass this in?
` updateActivityFeedback: (activityId: string, activityIndex: number, platformStudentId: string, feedback: any) => void;`

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+                </div>
+                {this.renderRatings(crit)}
+              </div>
+            )}
+          </div>
+        </div>
+      );
+    }
+
+    return null;
```
mklewandowski, line 39: I would write this as ternary instead of using 2 returns.  But looking more closely at it, I think the better option is to simply not render this component in the parent if activityStarted is false.  So remove activityStarted from the props, take out the if(activityStarted) and take out the return null.  Then simply have the parent say:
`{ activityStarted && <RubricTable/> }`

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+        </div>
+          <div className={css.rubricDescriptionTitle}>{rubric.criteriaLabel}</div>
+        </div>
+        { rubric.ratings.map((rating: any) =>
+          <div className={css.rubricScoreHeader} key={rating.id}>
+            <div className={css.rubricScoreLevel}>{rating.label}</div>
+            {rubric.scoreUsingPoints && <div className={css.rubricScoreNumber}>({rating.score})</div>}
+          </div>
+        )
+        }
```
mklewandowski, line 61: you might consider comparing this to the map in line 26 so they are formatted the same to make it easier to read.  To match the map in line 26, it would look like this:
```
        {rubric.ratings.map((rating: any) =>
          <div className={css.rubricScoreHeader} key={rating.id}>
            <div className={css.rubricScoreLevel}>{rating.label}</div>
            {rubric.scoreUsingPoints && <div className={css.rubricScoreNumber}>({rating.score})</div>}
          </div>
        )}
```

mklewandowski: Not a huge deal, but it improves the readability to have these maps use the same spacing, layout etc.

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+    const { referenceURL } = rubric;
+    return (
+      <div className={css.columnHeaders}>
+        <div className={css.rubricDescriptionHeader}>
+          <div className={css.scoringGuideArea}>
+            <a className={css.launchButton}>
+              <LaunchIcon className={css.icon} href={referenceURL} target="_blank" />
+            </a>
+            Scoring Guide
+        </div>
```
mklewandowski, line 52: tab this div in for alignment

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+        }
+      </div>
+    );
+  }
+
+  private renderRatings = (crit: any) => {
+    const { rubric } = this.props;
+    const { ratings } = rubric;
+    return (
+      <div id={`${crit.id}_ratings`} className={css.ratingsGroup}>
```
mklewandowski, line 70: Is this `id` used?

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+          </div>
+        )
+        }
+      </div>
+    );
+  }
+
+  private renderRatings = (crit: any) => {
+    const { rubric } = this.props;
+    const { ratings } = rubric;
```
mklewandowski, line 68: You don't need to pass `crit` into this function,  You can get it here:
`const { criteria, ratings } = rubric;`

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+    return (
+      <div id={`${crit.id}_ratings`} className={css.ratingsGroup}>
+        {ratings.map((rating: any, index: number) => this.renderStudentRating(crit, rating, index))}
+      </div>
+    );
+  }
+
+  private renderStudentRating = (crit: any, rating: any, buttonIndex: number) => {
+    const { rubricFeedback } = this.props;
```
mklewandowski, line 77: Same here.  `crit` doesn't need to be passed.  Can get from props:
```
const { rubric, rubricFeedback } = this.props;
const { criteria } = rubric;
```

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+    const radioButtonKey = `${critId}-${ratingId}`;
+    const selected = (rubricFeedback && rubricFeedback[critId] && rubricFeedback[critId].id === ratingId);
+    // Tooltips displayed to teacher should actually show student description if it's available.
+    const ratingDescription =
+      (crit.ratingDescriptionsForStudent && crit.ratingDescriptionsForStudent[ratingId]) ||
+      (crit.ratingDescriptions && crit.ratingDescriptions[ratingId]) ||
+      null;
+    const isApplicableRating = crit.nonApplicableRatings === undefined ||
+      crit.nonApplicableRatings.indexOf(ratingId) < 0;
+    const isNotApplicable = !isApplicableRating;
```
mklewandowski, line 90: You could eliminate the local variable by simply using either `isApplicableRating` or `!isApplicableRating` in the test below.

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    return (
+      <div className={css.rubricScoreBox} key={radioButtonKey}>
+        <div className={css.rubricButton} title={(isApplicableRating) ? ratingDescription : "Not Applicable"}>
+          { isNotApplicable
+            ? <span className={css.noRating}>N/A</span>
+            : this.renderButton(critId, selected,ratingId, buttonIndex)
```
mklewandowski, line 97: spacing:
`: this.renderButton(critId, selected, ratingId, buttonIndex)`

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    const critId = crit.id;
+    const ratingId = rating.id;
+    const radioButtonKey = `${critId}-${ratingId}`;
+    const selected = (rubricFeedback && rubricFeedback[critId] && rubricFeedback[critId].id === ratingId);
+    // Tooltips displayed to teacher should actually show student description if it's available.
+    const ratingDescription =
+      (crit.ratingDescriptionsForStudent && crit.ratingDescriptionsForStudent[ratingId]) ||
+      (crit.ratingDescriptions && crit.ratingDescriptions[ratingId]) ||
+      null;
```
mklewandowski, line 87: What happens if null is put in the title below?

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  private rubricChange = (rubricFeedback: any, studentId: string) => {
+    const { rubric, activityId, activityIndex, updateActivityFeedback } = this.props;
+    let numFeedback = 0;
+    rubric.criteria.map((crit: any)=>{
+      rubricFeedback[crit.id].id !== "" && numFeedback++;
+    });
+
+    const hasBeenReviewed  = numFeedback !== 0 ? true : false;
+    activityId && studentId && updateActivityFeedback && updateActivityFeedback(activityId, activityIndex, studentId, { rubricFeedback, hasBeenReviewed } );
```
mklewandowski, line 152: I would break this into multiple lines.  That said, if updateActivityFeedback isn't optional in the props list you can remove that check:
`    activityId && studentId && updateActivityFeedback(activityId, activityIndex, studentId, { rubricFeedback, hasBeenReviewed });`

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+      </button>
+    );
+  }
+
+  private rubricChange = (rubricFeedback: any, studentId: string) => {
+    const { rubric, activityId, activityIndex, updateActivityFeedback } = this.props;
+    let numFeedback = 0;
+    rubric.criteria.map((crit: any)=>{
+      rubricFeedback[crit.id].id !== "" && numFeedback++;
+    });
```
mklewandowski, line 149: I probably would not use a map here because your not mapping each array entry into some new value or output.  You're summing so a forEach might look cleaner.  Something like:
```
    rubric.criteria.forEach((crit: any) => {
      if (rubricFeedback[crit.id].id !== "") {
        numFeedback++
      }
    });
```

mklewandowski (changes requested):
Looking good.  I added comments for some code cleanup.  Should be pretty straightforward.

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,155 @@
...
+          }
+        </div>
+      </div>
+    );
+  }
+
+  private renderButton = (critId: string, selected: boolean, ratingId: string, buttonIndex: number) => {
+    const handleRatingChange = (buttonIndex: number) => () => {
+      const { rubric, rubricFeedback } = this.props;
+      const deselect = rubric.ratings.findIndex((r: any) => r.id === (rubricFeedback[critId].id)) === buttonIndex;
```
mklewandowski, line 107: Might need something like this to fix the error that @emcelroy noted:
`const deselect = rubric.ratings.findIndex((r: any) => r.id === (rubricFeedback?.[critId].id)) === buttonIndex;`

js/components/portal-dashboard/feedback/rubric-table.tsx
mklewandowski: Similar to the error that @emcelroy mentioned, you might need this as well:
`if (rubricFeedback[crit.id]?.id !== "") {`

mklewandowski: Basically we need some extra protection for when rubricFeedback doesn't have the entries yet.

js/components/portal-dashboard/feedback/rubric-table.tsx
eireland: Crit is actually the mapped criteria, so would have to map criteria at each function if I use the props.

js/components/portal-dashboard/feedback/rubric-table.tsx
eireland: same as above

js/components/portal-dashboard/feedback/rubric-table.tsx
eireland: Not sure.

eireland: not sure

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,154 @@
...
+      </button>
+    );
+  }
+
+  private rubricChange = (rubricFeedback: any, studentId: string) => {
+    const { rubric, activityId, activityIndex, updateActivityFeedback } = this.props;
+    let numFeedback = 0;
+    rubric.criteria.forEach((crit: any) => {
+      if (rubricFeedback && rubricFeedback[crit.id]) {
+        if  (rubricFeedback[crit.id].id !== "") {
```
mklewandowski, line 144: extra space sneaked in here

js/components/portal-dashboard/feedback/rubric-table.tsx
```diff
@@ -0,0 +1,154 @@
...
+          numFeedback++;
+        }
+      }
+    });
+
+    const hasBeenReviewed  = numFeedback !== 0 ? true : false;
```
mklewandowski, line 150: Can be simplified to:
`const hasBeenReviewed  = numFeedback !== 0`;

mklewandowski (approved):
👍 Looks good!  I added a couple of final comments for code tweaks:
- extra spacing in if statement
- can simplify expression (don't need ternary)

cypress/integration/portal-dashboard/portal-dashboard-feedback.spec.js
```diff
@@ -86,7 +86,7 @@ context("Portal Dashboard Feedback Panel", () => {
     });
   });
   context('Feedback Rows', () => {
-    describe('verify activity-level feedback textareas appear and accept input', () => {
+    describe('verify activity-level feedback appear and accept input', () => {
```
emcelroy, line 89: This is a very picky comment on a tiny typo, but I think it'd be good to change "appear" to "appears." Or maybe say something like "feedback elements" appear? Not critical though.

emcelroy (approved):
Looks good 👍  I left one non-critical suggestion in a comment.

PR #302, opened 2021-01-26
Response detail/Feedback View horizontal scroll

eireland:
Adds a div wrapper for response details elements to add horizontal scroll when feedback view is shown


mklewandowski (commented):
👍 I only tested briefly, but this looks good!

PR #303, opened 2021-01-26
Wraps legend to two lines if viewport is less than 1000px

eireland:
Picked 1000px viewable area because that's when legend starts getting cut off.

### DIFF
--- a/css/portal-dashboard/progress-view-legend.less
+++ b/css/portal-dashboard/progress-view-legend.less
@@ -86,7 +86,7 @@
     flex-wrap: wrap;
     width: 60%;
     min-width: 450px;
-    left: calc(((100vw - @left-panel-width) / 2) - (@legend-width/3));
+    left: calc(((100vw - @left-panel-width) / 2) - 225px);
   }
   .legendKey {
     flex-wrap: nowrap;
--- a/css/portal-dashboard/progress-view-legend.less
+++ b/css/portal-dashboard/progress-view-legend.less
@@ -84,8 +84,7 @@
     bottom: -30px;
     justify-content: center;
     flex-wrap: wrap;
-    width: 60%;
-    min-width: 450px;
+    width: 450px;
     left: calc(((100vw - @left-panel-width) / 2) - 225px);
   }
   .legendKey {
--- a/css/portal-dashboard/progress-view-legend.less
+++ b/css/portal-dashboard/progress-view-legend.less
@@ -11,7 +11,6 @@
   box-sizing: border-box;
   border-radius: 16px;
   width: @legend-width;
-  height: 32px;
   line-height: 32px;
   background-color: white;
   left: calc(((100vw - @left-panel-width) / 2) - @half-width);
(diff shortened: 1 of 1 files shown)


mklewandowski (approved):
👍 Looks good.  I would have though that we could do this entirely with flex boxes (that's what they're designed to do - wrap content to multiple lines), but not sure that's worth the time to invest in that approach.  Otherwise, I would at least update that left position.  I think my change makes it a little better.

css/portal-dashboard/progress-view-legend.less
```diff
@@ -79,3 +78,24 @@
...
 }
+@media screen and (max-width: 1000px) {
+  .legendContainer {
+    line-height: 25px;
+    bottom: -30px;
+    justify-content: center;
+    flex-wrap: wrap;
+    width: 60%;
+    min-width: 450px;
+    left: calc(((100vw - @left-panel-width) / 2) - @left-panel-width);
```
mklewandowski, line 89: I would think that you don't need 100vw to get a left position. Isn't it a constant?

css/portal-dashboard/progress-view-legend.less
```diff
@@ -79,3 +78,24 @@
     }
   }
 }
+@media screen and (max-width: 1000px) {
+  .legendContainer {
+    line-height: 25px;
+    bottom: -30px;
+    justify-content: center;
+    flex-wrap: wrap;
+    width: 60%;
```
mklewandowski, line 87: Do we really mean to use 60% when other pieces are static?  Seems like it shouldn't be a percentage.

mklewandowski: Seems to work though.

css/portal-dashboard/progress-view-legend.less
mklewandowski: Or is this to center it?

mklewandowski: In which case you'll get a better result with:
`left: calc(((100vw - 240px) / 2) - 225px);`

mklewandowski: since it's basically, ((screenWidth - leftPanelWidth) / 2) - (legendWidth / 2)
legendWidth / 2 is not the same as leftPanelWidth.

eireland: It's to center it. Changing it so to legendWidth/3 seems to work better than legendWidth/2

mklewandowski: That shouldn't be the case.  Why divide it by 3?
Did you try `left: calc(((100vw - 240px) / 2) - 225px);`?

mklewandowski: Here is using my calculation:
image

mklewandowski: Dividing by 3 will move it too far to the right.  legendWidth is 450px so 225px is what should be used.

mklewandowski: Again, the standard mechanism for this type of centering when an object has left edge positioning is to take availableWidth / 2 - objectWidth / 2.

mklewandowski: Using that we get perfect centering:
center

mklewandowski (commented):
I think we should table this PR for now in favor of this one:
https://github.com/concord-consortium/portal-report/pull/304

PR #304, opened 2021-01-27
Some content inaccessible in narrow viewports

emcelroy:
PT Stories: 

https://www.pivotaltracker.com/story/show/176649578
https://www.pivotaltracker.com/story/show/176648006

[#176649578]
[#176648006]

* Allow user to scroll horizontally to view content that was being cut off.
* Prevents progress view legend from being overlapped by other elements.


mklewandowski (approved):
👍 I think this is an improvement on the current behavior.  Ultimately I think we need a responsive design, but that is far outside the scope of this PR.

PR #305, opened 2021-01-27
176599991 feedback sort

mklewandowski:
This PR splits the feedback sort and progress/response sort into two unique pieces of Redux state.  This allows us to keep two different sort lists that are displayed to the user.


eireland (approved):
Looks good. Merge it!

PR #307, opened 2021-01-28
Change log manager url based on portal data [#176704163]

dougmartin:


scytacki (approved):
seems good to me.
I'm guessing that it will need a test to keep the coverage level up.
You might be able to use nock to see if a log message is posted to the correct location based on some mocked rawPortalData.

dougmartin:
@scytacki I added commits to fix an unescaped dot in the regex and to increase code coverage over logging.

scytacki (approved):
Looks great to me.

PR #311, opened 2021-01-29
Adds the refresh sort button to pop-class-nav

eireland:
Part of the bigger feedback sorting story. This just adds the refresh sort button. It doesn't do anything yet.

### DIFF
--- a/css/portal-dashboard/response-details/popup-class-nav.less
+++ b/css/portal-dashboard/response-details/popup-class-nav.less
@@ -144,7 +144,6 @@
       flex-flow: row;
       align-items: center;
       margin: 10px 0 0 20px;
-      cursor: pointer;
 
       .refreshButton {
         height: 30px;
@@ -153,6 +152,8 @@
         border: solid 1.5px @cc-charcoal-light1;
         border-radius: 5px;
         align-items: center;
+        cursor: pointer;
+
         &:hover {
           background-color: @cc-teal-light4;
         }
--- a/css/portal-dashboard/response-details/popup-class-nav.less
+++ b/css/portal-dashboard/response-details/popup-class-nav.less
@@ -87,7 +87,7 @@
     }
   }
   .popupClassNavControllers {
-    height: 183px;
+    height: 200px;
     padding-top: 10px;
 
     .spotlightToggle {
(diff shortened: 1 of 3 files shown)


js/containers/portal-dashboard/popup-class-nav.tsx
```diff
@@ -174,6 +176,17 @@ class PopupClassNav extends React.PureComponent<IProps>{
     );
   }
 
+  private renderRefreshButton = () => {
+    return (
+      <div className={css.refreshSortContainer}>
+        <button className={`${css.refreshButton}`}>
```
mklewandowski, line 182: Can be simplified to `<button className={css.refreshButton}>`

css/portal-dashboard/response-details/popup-class-nav.less
```diff
@@ -138,5 +138,50 @@
         color: @cc-charcoal;
       }
     }
+
+    .refreshSortContainer {
+      display: flex;
+      flex-flow: row;
+      align-items: center;
+      margin: 10px 0 0 20px;
+      cursor: pointer;
```
mklewandowski, line 147: I don't think you want this cursor on the container, but rather on the button.  Otherwise you get the weird affect where the cursor changes to the finger on the text (which isn't clickable) but not the button itself.  So delete `cursor: pointer;` above....

css/portal-dashboard/response-details/popup-class-nav.less
```diff
@@ -138,5 +138,50 @@
...
+
+      .refreshButton {
+        height: 30px;
+        width: 30px;
+        background-color: @cc-teal-light6;
+        border: solid 1.5px @cc-charcoal-light1;
+        border-radius: 5px;
+        align-items: center;
```
mklewandowski, line 155: ...and then add `cursor: pointer;` here inside the `refreshButton` class.

mklewandowski (approved):
👍 Looks good.  We need to tweak the `cursor: pointer;` CSS (per my comment, move it from the `refreshSortContainer` class and inside of the `refreshButton` class instead).  And there is one line where you can simplify the CSS string.  However, I'm going to approve in case you want to make the changes and merge tonight or this weekend and the approval holds up your work.

PR #312, opened 2021-02-01
Fix bad regex in resource url parser

dougmartin:
The regex was returning the last id digit instead of all the id digits.  This wasn't caught in the test coverage as the tests all used single digit ids.


scytacki (approved):
Looks good to me.

PR #313, opened 2021-02-02
Add logging username to report state

dougmartin:
Adds the username in the form of `<userid>@<domain>` to the report state so that it can used when logging


scytacki (approved):
looks good to me.

I wish our user naming was more consistent but now isn't the time to fix it.

PR #314, opened 2021-02-02
Awaiting Feedback Refresh button state

mklewandowski:
This PR updates the enabled/disabled state of the awaiting feedback refresh button.  State is maintained in Redux and passed to components as needed.  Changes include:
- add Redux state to keep track of refresh button enabled state
- when refresh button is pressed, disable button
- when feedback is changed, enable button
- when change between awaiting feedback and student name sort, disable button
- hide button if in student name sort mode

The next step will be to have the button click update the sort.  This will require feedbacks keeping track of if they have been modified since the last sort.

### DIFF
--- a/js/containers/portal-dashboard/popup-class-nav.tsx
+++ b/js/containers/portal-dashboard/popup-class-nav.tsx
@@ -56,7 +56,6 @@ class PopupClassNav extends React.PureComponent<IProps>{
                            ? "Awaiting feedback"
                            : listViewMode === "Question" ? "Questions" : "Class";
     const containerLabelType = viewMode === "FeedbackReport" || listViewMode === "Question" ? undefined : "students";
-
     return (
       <div className={`${css.popupClassNav} ${css.column}`}>
         {this.renderViewListOptions()}
@@ -65,7 +64,8 @@ class PopupClassNav extends React.PureComponent<IProps>{
           <CountContainer numItems={numItems} containerLabel={containerLabel} containerLabelType={containerLabelType} />
           {this.renderSortMenu()}
           {listViewMode === "Student" && viewMode === "ResponseDetails" && this.renderSpotlightToggle()}
-          {viewMode === "FeedbackReport" && feedbackSortByMethod === SORT_BY_FEEDBACK_PROGRESS && this.renderRefreshButton()}
+          {viewMode === "FeedbackReport" && listViewMode === "Student" && feedbackSortByMethod === SORT_BY_FEEDBACK_PROGRESS &&
+           this.renderRefreshButton()}
         </div>
       </div>
     );
--- a/js/containers/portal-dashboard/popup-class-nav.tsx
+++ b/js/containers/portal-dashboard/popup-class-nav.tsx
@@ -47,7 +47,8 @@ class PopupClassNav extends React.PureComponent<IProps>{
   }
 
   render() {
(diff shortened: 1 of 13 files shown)


js/containers/portal-dashboard/popup-class-nav.tsx
```diff
@@ -4,31 +4,34 @@ import { AnonymizeStudents } from "../../components/portal-dashboard/anonymize-s
 import { CustomSelect, SelectItem } from "../../components/portal-dashboard/custom-select";
 import { DashboardViewMode, FeedbackLevel, ListViewMode } from "../../util/misc";
 import { makeGetStudentFeedbacks } from "../../selectors/activity-feedback-selectors";
+import { getfeedbackSortRefreshEnabled } from "../../selectors/dashboard-selectors";
 import { CountContainer } from "../../components/portal-dashboard/count-container";
 import { SORT_BY_NAME, SORT_BY_MOST_PROGRESS, SORT_BY_LEAST_PROGRESS, SORT_BY_FEEDBACK_NAME,
-         SORT_BY_FEEDBACK_PROGRESS } from "../../actions/dashboard";
+         SORT_BY_FEEDBACK_PROGRESS, setFeedbackSortRefreshEnabled } from "../../actions/dashboard";
```
emcelroy, line 10: I would've thought this should be `SET_FEEDBACK_SORT_REFRESH_ENABLED`. Is this correct though?

emcelroy (approved):
I think this looks good generally 👍 I did leave one question in an inline comment.

Also, I don't know if it matters at this point, but when I switch to view list by question, the refresh button is visible even though the sort menu essentially goes away. Is it supposed to be visible there?

js/containers/portal-dashboard/popup-class-nav.tsx
mklewandowski: I think it's correct to import `setFeedbackSortRefreshEnabled ` - it's the actual function that we need to call.  The other ones are the sort actions themselves that we tie directly to the dropdown menu.  Looking at it though, it seems like a weird route that was originally designed to handle the sorts.  Oh well - there is probably some logic behind it that I am not familiar with.

mklewandowski:
> I think this looks good generally 👍 I did leave one question in an inline comment.
> 
> Also, I don't know if it matters at this point, but when I switch to view list by question, the refresh button is visible even though the sort menu essentially goes away. Is it supposed to be visible there?

I think you're right that the button should be hidden when looking at question level feedback.  I will make that change.

mklewandowski:
gm


> > I think this looks good generally 👍 I did leave one question in an inline comment.
> > Also, I don't know if it matters at this point, but when I switch to view list by question, the refresh button is visible even though the sort menu essentially goes away. Is it supposed to be visible there?
> 
> I think you're right that the button should be hidden when looking at question level feedback. I will make that change.

Actually, I wonder if we're supposed to add awaiting feedback to that menu and then enable the menu.  I better check with Kiley.

mklewandowski:
> I think this looks good generally 👍 I did leave one question in an inline comment.
> 
> Also, I don't know if it matters at this point, but when I switch to view list by question, the refresh button is visible even though the sort menu essentially goes away. Is it supposed to be visible there?

After discussion, I went ahead and hit the refresh button in this case.  Thanks for catching.

PR #315, opened 2021-02-03
176492325 feedback fixes

mklewandowski:
This PR adds the following changes and fixes to the activity feedback sort:
- display sorted activity feedback when awaiting feedback sort is used
- fix bug in activity awaiting feedback sort where all activities were checked for feedback, need to use current or first
- fix bug in activity awaiting feedback sort where unstarted activities were not sorted properly
- update refresh button state when rubric is selected
- update container height
NOTE that with these changes as soon as activity feedback is entered, the feedback will be sorted to the bottom of the list. This will be fixed in upcoming work.

### DIFF
--- a/js/selectors/dashboard-selectors.js
+++ b/js/selectors/dashboard-selectors.js
@@ -5,6 +5,10 @@ import { SORT_BY_NAME, SORT_BY_MOST_PROGRESS, SORT_BY_LEAST_PROGRESS,
 import { compareStudentsByName } from "../util/misc";
 import { fromJS } from "immutable";
 
+const kSortGroupFirst = 1;
+const kSortGroupSecond = 2;
+const kSortGroupThird = 3;
+
 // Inputs
 const getActivities = state => state.getIn(["report", "activities"]);
 export const getAnonymous = state => state.getIn(["report", "anonymous"]);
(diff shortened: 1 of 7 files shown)


js/selectors/dashboard-selectors.js
```diff
@@ -5,6 +5,10 @@ import { SORT_BY_NAME, SORT_BY_MOST_PROGRESS, SORT_BY_LEAST_PROGRESS,
 import { compareStudentsByName } from "../util/misc";
 import { fromJS } from "immutable";
 
+const kSortGroupFirst = 1;
+const kSortGroupSecond = 2;
+const kSortGroupThird = 3;
```
eireland, line 10: May be more helpful if the names were more descriptive like kSortGroupAwaitingFeedback, kSortGroupFeedbackGiven, and kSortGroupNotStarted.

eireland (approved):
Just that one suggestion. Otherwise looks good.

js/selectors/dashboard-selectors.js
mklewandowski: I was hoping the opposite - that these could be reused for any type of sort and all that we need to do is put some data into a group.  This way when comparing we know first always comes before second, second before third, etc.  But if I have names like `kSortGroupAwaitingFeedback` it doesn't really imply any sort meaning and it's up to someone reading the code to make sense of where it gets positioned.  Basically I'm trying to reduce cognitive load and make this sorting as easy as possible for engineers.

emcelroy (approved):
This looks good and I see you've already merged it. I was going to note though that I expected the the awaiting feedback sort to not only put the students awaiting feedback at the top of the list, but to then also sort that sub group of students alphabetically by last name. I don't think that's in the spec. But maybe it was discussed offline and the consensus was it didn't matter? It seems like the natural thing to expect. I can imagine teachers looking specifically for a certain student and the alphabetized order helping with that. I don't think it's crucial though, and if others have said it's not important, their opinion probably matters more than mine.

mklewandowski:
> This looks good and I see you've already merged it. I was going to note though that I expected the the awaiting feedback sort to not only put the students awaiting feedback at the top of the list, but to then also sort that sub group of students alphabetically by last name. I don't think that's in the spec. But maybe it was discussed offline and the consensus was it didn't matter? It seems like the natural thing to expect. I can imagine teachers looking specifically for a certain student and the alphabetized order helping with that. I don't think it's crucial though, and if others have said it's not important, their opinion probably matters more than mine.

Yes, I think this is a good idea and is something that @eireland could look into when finishing this work.  My main goal here was to get us over the hump with this work.  I'm punting on the nuance since I have limited time to work on this project.  But I agree that this should be added to our cleanup list for feedback sorting.

PR #316, opened 2021-02-03
176492325 question sort refresh button

mklewandowski:
This PR updates the awaiting feedback search.  Changes include:
- add question sort, implemented in `getQuestionFeedbackSortedStudents`
- when refresh button is pressed or sort is changed, mark existing feedback since last sort with `existingFeedbackSinceLastSort` flag so sort knows which feedback has been added since the last sort (and can treat the feedback as if it were empty)
- improve handling of empty string feedbacks (most likely caused by deletion)
- fix Cypress tests


emcelroy (commented):
Looks good 👍

emcelroy (approved):
Looks even better! That issue we discussed on Slack is gone now.

mklewandowski:
> Looks even better! That issue we discussed on Slack is gone now.

There's still one feedback badge issue that I'm looking into.  I'll merge soon.

mklewandowski:
> > Looks even better! That issue we discussed on Slack is gone now.
> 
> There's still one feedback badge issue that I'm looking into. I'll merge soon.

I think it's outside the scope of this work.  I'm going to merge.

PR #320, opened 2021-02-09
176648006 legend wrap

eireland:
Re-opens the legend wrapping in narrow screen widths


mklewandowski (commented):
@eireland try removing the `left` calculation and setting `bottom` to -5px.

PR #322, opened 2021-02-11
Fix double scroll

eireland:


mklewandowski (changes requested):
When I run the branch and make it narrow on my system, I see this:
image

I admit I'm not entirely sure how to repro the original bug though.

mklewandowski:
Maybe I'm not understanding the bug.  Is it the double scroll or is it a z order issue caused by studentList?  Or is it both?  I would try:
- reverting out all of Ethan's changes from the commit that I linked
- add a z-index to studentList class so it appears above the activity summary content

eireland:
The PR got bigger. In order to fix the double scroll, I had to redefine the areas so that `overflow-x: hidden` style can be applied to the correct area. The side efect was to fix the bug where two headers exists when in response detail or feedback view. Also removed the min widths so we can get smaller before we can't access some of the right most elements.

mklewandowski:
I'm till struggling to understand what the bug is.  Is it that they don't like the double scrollbars or is it the overlapping content?

If it's the overlapping content, then I think the fix is just one line.  Adding `z-index: 1;` to the studentList class.  Try this branch locally:
https://github.com/concord-consortium/portal-report/tree/zorder-test
image


If the bug is that they don't like the double scroll bars, then we should talk about what the expected behavior is intended to be.

eireland:
The bug was the double scroll bar which I tried to fix by doing `overflow-x: hidden`, which caused the above bug with the `<AnswerCompact>` overlapping the student names. Adding `z-index: 1` to the student names causes the sort dropdown to go behind it. If you make class-nav a higher z-index, the view dropdown goes behind it. 
Then the double scroll bar was also happening in the response/feedback view, so made the overflow-x hidden there as well, but a lot of things because inaccessible, so fiddled around with the flexiness. 
And we are rendering the header twice when we are in response/feedback view. So because the classNav z-index is higher, the view dropdown in the second header gets hidden behind it. Increasing the-z-index was making the two headers act weird. So I refactored the main container to just show one header so we have more control on how it's shown regardless of view.

mklewandowski:
> The bug was the double scroll bar which I tried to fix by doing `overflow-x: hidden`, which caused the above bug with the `<AnswerCompact>` overlapping the student names. Adding `z-index: 1` to the student names causes the sort dropdown to go behind it. If you make class-nav a higher z-index, the view dropdown goes behind it.
> Then the double scroll bar was also happening in the response/feedback view, so made the overflow-x hidden there as well, but a lot of things because inaccessible, so fiddled around with the flexiness.
> And we are rendering the header twice when we are in response/feedback view. So because the classNav z-index is higher, the view dropdown in the second header gets hidden behind it. Increasing the-z-index was making the two headers act weird. So I refactored the main container to just show one header so we have more control on how it's shown regardless of view.

But AnswerCompact overlapping the student names is a bug on master.  You're right that it takes more than a single z-index. But can't we fix it by simply setting studentList zindex to 1 and classNav zindex to 2?  2 lines of CSS?

I still have the following question:
- the double scroll was added to access all content.  Is it a bug?  Don't we need it to access that content?  I again think we should clarify what the bug is.  If we remove that double scroll then users cannot access all content.  Is that what Kiley and Trudi want?

We should add some clarification around this PR.  I would add a description so it is clear what this PR is doing.

css/portal-dashboard/header.less
```diff
@@ -8,7 +8,6 @@
   box-sizing: border-box;
   height: @header-height;
   margin-bottom: 3px;
-  min-width: 1100px;
```
emcelroy, line 11: If I remember correctly, I added these `min-width` values so if your browser window was narrow, you could still horizontally scroll over to view content that wasn't visible. With these changes you can't do that anymore. Any content on the right side of the dashboard that's not visible in the browser window becomes inaccessible unless you can resize the browser.

Was it decided that that's OK though?

mklewandowski: I agree.  I think we should clarify with Kiley and Trudi what they want changed/fixed.  I posted in the Slack thread to get clarification.

mklewandowski: We definitely want to fix that overlap bug, but I think that is just a couple lines of CSS which could be extracted into a much simpler PR.

eireland:
I'll pull out the min-width and overflow-x changes I made, and keep the fix for the overlapping answer compact.
This means the z-index and the header refactor changes will stay.

eireland:
Closing this pull request to split the two work and reopening with correct branch name

PR #324, opened 2021-02-17
Changes the z-index of the different parts of the progress view.

eireland:
Refactors the header to only show one header when in feedback and response detail view.
Updates cypress test to reflect the changes.

### DIFF
--- a/js/containers/portal-dashboard/portal-dashboard-app.tsx
+++ b/js/containers/portal-dashboard/portal-dashboard-app.tsx
@@ -169,78 +169,79 @@ class PortalDashboardApp extends React.PureComponent<IProps, IState> {
 
     return (
       <div className={css.portalDashboardApp}>
-        {sequenceTree && this.renderHeader(assignmentName, "ProgressDashboard" )}
+
+        {this.renderHeader(assignmentName, viewMode )}
         {activityTrees &&
-          <div>
-            <div className={css.navigation}>
-              <ClassNav
-                anonymous={anonymous}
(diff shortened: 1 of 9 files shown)


js/containers/portal-dashboard/portal-dashboard-app.tsx
```diff
@@ -279,22 +279,23 @@ class PortalDashboardApp extends React.PureComponent<IProps, IState> {
...
-        userName={userName}
-        setCompact={headerViewMode === "ProgressDashboard" ? setCompactReport : undefined}
-        setHideFeedbackBadges={headerViewMode === "ProgressDashboard" ? setHideFeedbackBadges : undefined}
-        assignmentName={assignmentName}
-        trackEvent={trackEvent}
-        setDashboardViewMode={this.setDashboardViewMode}
-        viewMode={viewMode}
-        colorTheme={color}
-      />
+      sequenceTree && <Header
```
emcelroy, line 289: I'm not 100% sure about this particular instance, but I think if what comes after `&&` is made up of multiple lines, our convention is to have it start on a new line, as in line 175 above, right?

emcelroy (approved):
This looks good and worked as expected when I tested it out. I just left one inline comment/question about how one block of code is formatted.

mklewandowski:
What were the changes in the portal dash app for?  When I tested locally last week, I only needed two lines of CSS changes to fix the z-order.

eireland:
Adding the different z-index was affecting the z-index in the response/feedback view.

PR #325, opened 2021-02-24
Fix activity level feedback view when there is no rubric

eireland:
Bug was when user clicked on Activity toggle in feedback view, the page would turn blank. This was because it would fail when there was no rubric feedback.
This PR adds a check to see if there is rubric feedback before trying to check if rubric feedback had been given.


emcelroy (approved):
Looks good to me 👍

mklewandowski (approved):
👍

PR #329, opened 2021-03-02
Removes second colon in class-nav

eireland:
Removes a redundant colon from class nav student label
Screen Shot 2021-03-02 at 10 20 13 AM


emcelroy (approved):
👍

PR #333, opened 2021-04-29
Order map keys when generating answer hash to ensure repeatable hash results

mklewandowski:
This PR fixes a bug that caused valid image question feedback to sporadically be hidden.  This was caused by a mismatch between the answer hash stored with the feedback and the answer hash created in the dashboard from the current image question answer - this mismatch caused us to hide the feedback (thinking that it was out of date because the answer had changed).  Even when the user-facing image question answer has not changed, the stored structure of the answer can change.  Most answers are simple text strings.  However, the image question answer is stored as an Immutable JS `Map` with two keys: `image` and `text`.  Depending on how that `Map` was created, we would sometimes get the `image` key first and sometimes get the `text` key first.  In other words, the key order was not predictable.  We then stringified the `Map`, ran it through the `md5` hash function, and saved the results.  But because the key order was not stable, we could actually get two different hash values from a single user-facing answer.  
The solution is to detect if we have an answer in the form of a `Map`, order the keys in alpha-numeric order, and then stringify the key-ordered structure.  This guarantees the same hash each time.  I don't think we need to go deeper than looking at the top-level answer itself since answers do not typically have much complexity (they are  usually a simple string - the image question was the only example I found so far that required anything beyond a string).  But I am willing to dig into looking for a `Map` within a `Map` if folks think that we will have answers with that complexity.

### DIFF
--- a/js/util/misc.ts
+++ b/js/util/misc.ts
@@ -27,7 +27,9 @@ export const compareStudentsByName = (student1: Map<string, any>, student2: Map<
 export const answerHash = (answer: Map<string, any>) => {
   let answerContent = answer.get("answer");
   if (typeof answerContent !== "string") {
-    answerContent = JSON.stringify(answerContent.toJS());
+    answerContent = Map.isMap(answerContent)
+      ? JSON.stringify(answerContent.sortBy((v: any, k: any) => k).toJS())
+      : JSON.stringify(answerContent.toJS());
   }
   return md5(answerContent);
 };
--- a/js/util/misc.ts
+++ b/js/util/misc.ts
@@ -27,9 +27,11 @@ export const compareStudentsByName = (student1: Map<string, any>, student2: Map<
 export const answerHash = (answer: Map<string, any>) => {
   let answerContent = answer.get("answer");
   if (typeof answerContent !== "string") {
+    // the non-Map case likely doesn't hit at present, but this adds protection
+    // in case future answerContent has non-string/non-Map value
     answerContent = Map.isMap(answerContent)
       ? JSON.stringify(answerContent.sortBy((v: any, k: any) => k).toJS())
-      : JSON.stringify(answerContent.toJS());
+      : JSON.stringify(answerContent);
   }
   return md5(answerContent);
 };
--- a/test/util/misc_spec.ts
+++ b/test/util/misc_spec.ts
@@ -1,0 +1,36 @@
+import { Map } from "immutable";
+import { answerHash } from "../../js/util/misc";
+
+describe("misc util functions", () => {
+  it("determines if answer hash is properly generated", () => {
+
+    const simpleAnswer1 = Map({ answer: "foo" });
+    const simpleAnswer2 = Map({ answer: "foo" });
+    const simpleAnswer3 = Map({ answer: "bar" });
+
+    const simpleAnswer1Hash = answerHash(simpleAnswer1);
+    const simpleAnswer2Hash = answerHash(simpleAnswer2);
+    const simpleAnswer3Hash = answerHash(simpleAnswer3);
+
+    expect(simpleAnswer1Hash).toEqual(simpleAnswer2Hash);
+    expect(simpleAnswer1Hash).not.toEqual(simpleAnswer3Hash);
+    expect(simpleAnswer2Hash).not.toEqual(simpleAnswer3Hash);
+
+    const complexAnswerContent1 = Map({ image: "www.test.com/image.jpg", text: "foo" });
+    const complexAnswerContent2 = Map({ text: "foo", image: "www.test.com/image.jpg" });
+    const complexAnswerContent3 = Map({ image: "www.test.com/image.jpg", text: "bar" });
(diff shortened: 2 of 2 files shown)


js/util/misc.ts
```diff
@@ -27,7 +27,9 @@ export const compareStudentsByName = (student1: Map<string, any>, student2: Map<
 export const answerHash = (answer: Map<string, any>) => {
   let answerContent = answer.get("answer");
   if (typeof answerContent !== "string") {
-    answerContent = JSON.stringify(answerContent.toJS());
+    answerContent = Map.isMap(answerContent)
+      ? JSON.stringify(answerContent.sortBy((v: any, k: any) => k).toJS())
+      : JSON.stringify(answerContent.toJS());
```
scytacki, line 32: @mklewandowski can you add a test for this code?

I'd guess that this has been a problem for a long time now so feedback on image questions and probably generic interactives has had this problem for a while. We could try to do something fancy, to fix the old data, but I don't think it is worth it.

scytacki (changes requested):
It'd be useful to add some tests of this function.

js/util/misc.ts
kswenson: FYI, the npm library I mentioned in standup that handles this is [object-hash](https://www.npmjs.com/package/object-hash). In this case, you could use `answerContent = object-hash(answerContent.toJS())`, although that wouldn't be backwards-compatible with any existing hashes.

kswenson (commented):
I don't understand the equities here, so I simply left a comment.

scytacki:
@mklewandowski I'd be curious to see what the answerContent is for a table interactive. There is one in the demo portal-report code it is the Question #1 of the second activity. http://portal-report.concord.org/branch/master/

I'm curious because I wonder if its answerContent is also a Map if so it might be a deeply nested Map. Or maybe it is a single level map and the values of the keys are json strings or something. Or maybe the answerContent is some simple string and the actual state of the table interactive is available in a different field.

js/util/misc.ts
mklewandowski: @scytacki I added unit tests to test both the previous working case and this newly working case where the answer itself is a Map.

mklewandowski:
> @mklewandowski I'd be curious to see what the answerContent is for a table interactive. There is one in the demo portal-report code it is the Question #1 of the second activity. http://portal-report.concord.org/branch/master/
> 
> I'm curious because I wonder if its answerContent is also a Map if so it might be a deeply nested Map. Or maybe it is a single level map and the values of the keys are json strings or something. Or maybe the answerContent is some simple string and the actual state of the table interactive is available in a different field.

I'll take a look at this and see what I find.

mklewandowski:
> @mklewandowski I'd be curious to see what the answerContent is for a table interactive. There is one in the demo portal-report code it is the Question #1 of the second activity. http://portal-report.concord.org/branch/master/
> 
> I'm curious because I wonder if its answerContent is also a Map if so it might be a deeply nested Map. Or maybe it is a single level map and the values of the keys are json strings or something. Or maybe the answerContent is some simple string and the actual state of the table interactive is available in a different field.

Looks like the answer content (what we use for generating the hash) is actually a string in the table interactive case.  It's a big string that carries a lot of info (and looks like it was a JS Object that was converted to a string), but it's a string.  For example:
```
{"version":1,"mode":"report","authoredState":"{\"columns\":[{\"heading\":\"Read-only value\",\"readOnly\":true,\"average\":false,\"chart\":false,\"chartColor\":\"\"},{\"heading\":\"Test value\",\"readOnly\":false,\"average\":false,\"chart\":false,\"chartColor\":\"\"},{\"heading\":\"Value 1 (m)\",\"readOnly\":false,\"average\":false,\"chart\":true,\"chartColor\":\"\"},{\"heading\":\"Value 2 (s)\",\"readOnly\":false,\"average\":false,\"chart\":true,\"chartColor\":\"green\"}],\"labels\":[\"a\",\"b\",\"c\"],\"chartWidth\":295,\"chartHeight\":240,\"rowLines\":1,\"chartAvgs\":false}","interactiveState":"{\"data\":[[\"a\",\"1\",\"2\",\"3\"],[\"b\",\"4\",\"5\",\"6\"],[\"c\",\"7\",\"8\",\"9\"]]}"}
```

So it should be safe.  I'm going to go through some of the other sample cases and see if I see any other oddities.

mklewandowski:
The only other `answerContent` that I've found so far that has complexity beyond a string is the multiple choice answer.  But this is a Map with a single entry which looks to be an array of size 1 with the selected answer so that should be safe.  So far only the image question will have the key order issue.

scytacki (approved):
Yeah I think this is good. I'd guess it has to do with immutable JS. It is good to know that all other answerContent are just strings so shouldn't be broken by this.

scytacki:
I'll go ahead and merge this. @mklewandowski can you figure out what the next version of the portal-report should be and label the story appropriately?

With Kevin on vacation it might fall to @eireland or @tejal-shah to test this release.

PR #334, opened 2021-05-03
Fix question nav height

mklewandowski:
This PR fixes the question nav height in the response details view when there is only a single activity.

image


eireland (approved):
LGTM

PR #344, opened 2021-06-24
Teacher / Dashboard report open an interactive with its saved state from a link

eireland:
Teacher / Dashboard report should support a way for users to open an interactive with its saved state from a link.
Opens student's iframe Interactive from student's and teacher's report Open in new tab link
Also opens from student's Activity Player show My Work button.


kswenson (changes requested):
See comments

kswenson (approved):
👍 LGTM

PR #345, opened 2021-07-28
Fix AP Report not loading for Sequences

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178787916

[#178787916]


scytacki (approved):
Looks good to me.

PR #347, opened 2021-08-14
Add researcher support

scytacki:
This uses the new portal researcher JWT support,
and the new report-service researcher support.

It does this by passing a target_user_id parameter to the portal's firebase JWT endpoint.

And it now passes the studentId param into the firestore queries.
This is needed so the firestore rules can only allow access to a specific student
based on the JWT issued by the portal.
This might also speed up the per student report since it will limit the number of answers returned.


pjanik (approved):
Looks good 👍

PR #351, opened 2021-08-30
Handle errors in image question documents

scytacki:
The AP is occasionally saving image question documents that don't have an answer field. This was breaking the portal-report.

The changes in this PR handle this case by showing and error message for this particular student answer. The once case that isn't handled well is in the standard report image carousel. In that case with a broken answer just a blank image is shown. Because we are trying to move away from the standard report, and this should only happen in error cases it seems fine to leave this broken.  If a teacher is curious and then show the responses then they will see the error message.

Here is what the standard report looks like:
image

In the portal dashboard "progress dashboard"
image

In the portal dashboard "response details":
image

https://www.pivotaltracker.com/story/show/179406265

### DIFF
--- a/js/util/misc.ts
+++ b/js/util/misc.ts
@@ -25,8 +25,23 @@ export const compareStudentsByName = (student1: Map<string, any>, student2: Map<
   }
 };
 
+// A fake MD5 hash that we'll use whenever the answer field is
+const MD5_FOR_UNDEFINED = "0def0def0def0def0def0def0def0def";
+
 export const answerHash = (answer: Map<string, any>) => {
   let answerContent = answer.get("answer");
+  if (answerContent === undefined) {
+    // There are error cases where the answer property is not defined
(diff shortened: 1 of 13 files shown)


scytacki:
The code coverage is lower, not because of these changes, but because of the random scrolling issue. When cypress doesn't scroll the window then less code is covered, and in this case the window wasn't scrolled.

js/util/misc.ts
```diff
@@ -25,8 +25,23 @@ export const compareStudentsByName = (student1: Map<string, any>, student2: Map<
   }
 };
 
+// A fake MD5 hash that we'll use whenever the answer field is
```
eireland, line 28: `answer field is` "?" I'm assuming undefined as the variable is named?

eireland (approved):
Looks good

js/util/misc.ts
scytacki: good catch

mklewandowski (commented):
LGTM! 👍

PR #352, opened 2021-08-31
remove obsolete updateReportSettingsInPortal

mklewandowski:
This PR removes the function `updateReportSettingsInPortal` which is no longer used and was causing errors when accessing user feedback (a PUT request with an updated feedback object was being sent to an invalid URL causing a 404 error).  Feedback is uploaded to Firestore, but no longer needs to be uploaded to the portal.


scytacki (approved):
looks good to me

PR #354, opened 2021-09-03
Image question no prompt bug

eireland:
Fixes a portal dashboard crash when rendering an image question that doesn't have a drawing prompt or a question prompt.
Adds error message to image carousel in report when answer image url is missing.
image


scytacki (changes requested):
The code changes look good to me.
My only requested change is that there is a test for this.
I think the best approach is to add a question in the example data that has no prompt. And then the cypress tests can verify this no prompt question doesn't crash the report.

scytacki (changes requested):
This looks good. 
However I tested it locally and the test doesn't fail when I remove the code change.

Specifically I reverted the changes in answer-modal.tsx, and then ran the whole cypress suite. All of the tests passed.  I reverted the changes by commenting out the new lines and copying and pasting the deleted line.

scytacki (changes requested):
The cypress tests have a few failures now.

scytacki (approved):
Looks good to me. 

I had to fix some new code that got lost when the rebase was done. I still prefer that you merge instead of rebase the branch.

PR #355, opened 2021-09-07
move question icon logic to question-utils

scytacki:
This also fixes an issue where the question has an unknown type.
in that case it uses the iframe_interactive icon.

There are couple of logic changes here that I'm not sure about:
- in answer-modal, it was using the answer's questionType, the types seem to imply that sometimes the question can be undefined. So this might be why the answer's questionType is used. However it makes no sense to me that there'd be an answer without a question. So this case of an undefined question seems like an error edge condition. I didn't try to track down when that case happens, so possible it is a common case and should be handled better
- in spotlight-student-list-dialog, the fact that the question is scored or not, was being ignored when looking up the icon. I assumed this was an oversight, but perhaps it was on purpose.


mklewandowski (approved):
👍 This looks good to me!

scytacki:
I'm going to merge this then. I'm going to open a new PR with some improvements to the code coverage. Most of the decrease in code coverage in this PR is because it removed several covered lines from files that aren't 100% covered.

PR #356, opened 2021-09-09
increase code coverage

scytacki:
This adds coverage to the dashboard:
- completed icon
- invalid answer type
- spotlight color changing button
- clicking on button to open the activity related to a question

It also attempt to fix an issue with the coverage fluctuating up and down when the dashboard (not portal-dashboard) would sometimes scroll horizontally and sometimes not scroll. 

To be able to test the completed icon I refactored the progress icons,
so I could add a data-cy property with the type of the icon.
I also changed the test data so there was one student with a fully completed activity.

For the invalid answer type I also changed the test data.

For the opening an activity link, I attempted to not just click on the button, but also monitor the dispatch event to verify that the click would actually cause a log event. However the approach I tried for spying on the dispatch didn't work. The test did increase the coverage of click event though, so I left the TODO comment incase we come back to this.


mklewandowski (approved):
👍

PR #357, opened 2021-09-13
Fix attachments manager options helper

pjanik:
https://www.pivotaltracker.com/n/projects/2441249/stories/179540761

I can't run reports for anonymous users locally (not sure why, it works for logged-in students though, I need to fix it), but I'm 99% sure it's gonna fix the issue.

The second commit is unrelated, I just ran the linter and noticed that it was complaining about quite a few errors.


scytacki (approved):
looks good to me

PR #361, opened 2021-09-18
Handle all kind of questions in iframe-standalone-app

pjanik:
[#179631327]

It's simple and it'll make Sharing Plugin more usable - students will be able to share any kind of question.


scytacki (approved):
I made some changes to get this fully working with multiple choice questions, and increase the test coverage

PR #363, opened 2021-09-22
Added support for anonymous LARA users [#179623345]

dougmartin:
- Added check for sourceKey
- Updated answer map generation to fall back to runKey when platformUserId is missing
- Hid controls in anonymous view

### DIFF
--- a/js/actions/index.ts
+++ b/js/actions/index.ts
@@ -11,7 +11,7 @@ import {
   reportQuestionFeedbacksFireStorePath,
   reportActivityFeedbacksFireStorePath,
   feedbackSettingsFirestorePath,
-  makeSourceKey
+  getSourceKey
 } from "../api";
 import {
   API_UPDATE_QUESTION_FEEDBACK,
@@ -75,8 +75,7 @@ export function fetchAndObserveData() {
   const runKeyValue = urlParam("runKey");
   if (runKeyValue) {
     const activity = urlParam("activity") || "";
-    const sourceKey = urlParam("sourceKey") || "";
-    const source = sourceKey ? sourceKey : (activity ? makeSourceKey(activity) : "");
+    const source = getSourceKey(activity);
     const answersSourceKey = urlParam("answersSourceKey") || source;
     return (dispatch: Dispatch, getState: any) => {
       dispatch({
--- a/js/actions/index.ts
+++ b/js/actions/index.ts
@@ -75,7 +75,8 @@ export function fetchAndObserveData() {
   const runKeyValue = urlParam("runKey");
   if (runKeyValue) {
     const activity = urlParam("activity") || "";
-    const source = activity ? makeSourceKey(activity) : "";
+    const sourceKey = urlParam("sourceKey") || "";
+    const source = sourceKey ? sourceKey : (activity ? makeSourceKey(activity) : "");
     const answersSourceKey = urlParam("answersSourceKey") || source;
     return (dispatch: Dispatch, getState: any) => {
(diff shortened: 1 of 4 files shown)


js/actions/index.ts
```diff
@@ -75,7 +75,8 @@ export function fetchAndObserveData() {
   const runKeyValue = urlParam("runKey");
   if (runKeyValue) {
     const activity = urlParam("activity") || "";
-    const source = activity ? makeSourceKey(activity) : "";
+    const sourceKey = urlParam("sourceKey") || "";
+    const source = sourceKey ? sourceKey : (activity ? makeSourceKey(activity) : "");
```
scytacki, line 79: Can you refactor the function https://github.com/concord-consortium/portal-report/blob/5355a12ee0f21b8b2fddbe58d56be00b6e5f343d/js/api.ts#L79-L83
So it takes a a url like `getSourceKey(url: string)`
Then you can use that function here and in the two other places where `getSourceKeyFromOffering` was used before.

js/reducers/report-reducer.ts
```diff
@@ -131,6 +131,7 @@ export default function report(state = new ReportState({}), action?: any) {
       .set("type", "student")
       .set("nowShowing", "student")
       .set("selectedStudentIds", Immutable.fromJS([action.runKey]))
+      .set("hideControls", true)
```
scytacki, line 134: I see now that hideControls is basically the same as `type: student`. Whenever we set the type to student we also have to set hideControls to true. It might be better to drop the hideControls property to simplify things, but what you did here is fine too.

js/selectors/report-tree.js
```diff
@@ -42,6 +42,9 @@ const isQuestionVisible = (question, featuredOnly) => {
   return true;
 };
 
+// allow for anonymous students
+const getStudentId = (answer) => answer.get("platformUserId") || answer.get("runKey");
```
scytacki, line 46: This makes sense, but instead of doing this, LARA should set the platform_user_id field with the value of the run_key when it saves the answers.  That is what the AP does. This way tools that are working with the firestore answers don't need to care as much about the run_key and can just treat the platform_user_id (plus platform_id if it is exists) as the unique identifier for the student.

Here is a working AP runKey based report:
https://portal-report.concord.org/branch/master/index.html?runKey=f671cd0c-75eb-4ceb-98ba-d3ea5ae3d10c&activity=https://authoring.staging.concord.org/activities/20874&resourceUrl=https://authoring.staging.concord.org/activities/20874&firebase-app=report-service-dev&sourceKey=authoring.staging.concord.org&answersSourceKey=activity-player.concord.org

This is what one of the answers looks like in firestore from in that report:
image

scytacki (changes requested):
A found a couple of things to fix.

js/reducers/report-reducer.ts
dougmartin: The `hideControls` is also used int he SelectionCheckbox component so I think it should be left as is otherwise that would need to be refactored.

scytacki:
@dougmartin just checking on this PR too. Did you want to refactor the getSourceKeyFromOffering function or are you going to leave it as is?

dougmartin:
@scytacki I had made that change after your review but forgot to push it.  I just did: 

https://github.com/concord-consortium/portal-report/pull/363/commits/32dde7b0ee8a1a3f4793675c47d8af797e9ef87e

scytacki (approved):
looks good to me.

PR #366, opened 2021-11-24
Updating production release instructions to include portal-report ver…

tejal-shah:
…sion update in activity-player


scytacki (approved):
looks good

scytacki:
One nuance to this is that if the AP is under active development, then we wouldn't want to just release the master branch. So if we need to do some urgent fix to the report when launched from the AP, then the process might involve finding the latest release tag of the AP, making a branch from this tag, and adding a PR that updates the portal report URL in this branch. 

I don't think it is worth complicating the documentation with all of those details though. Maybe just a note about checking what other changes are in master compared to the released version of the AP.   The release steps for the AP in the AP readme require doing this anyhow.

PR #367, opened 2021-12-08
fix crash when there is no feedback

eireland:
When feedback is turned on, opening the response details or feedback report crashes portal report when some students have not received feedback for a question. We were trying to trim a feedback that didn't exist. 
This adds a check to see if feedback existed before trying to trim it.


emcelroy (approved):
Looks good 👍

PR #368, opened 2021-12-09
Fix response details height (179810273)

mklewandowski:
This fixes the response details container height so that all of the student response and feedback rows can be accessed with scrollbars when the content doesn't fit vertically.

PT Story:
https://www.pivotaltracker.com/story/show/179810273

feedbacl-height


eireland:
Need to adjust the padding in `feedback-rows.less` `.feedback` to `10px 12px 26px` because of the scrollbar.
And just a nicety to match response detail view: In `response-details.less` `.feedbackRowsContainer`, add `margin-top: 3px;`

eireland (approved):
LGTM!

PR #369, opened 2021-12-10
Fallback to checking runKey for anonymous users when showing iframe previews [#180446631]

dougmartin:


scytacki (changes requested):
This looks good.

Is it easy to add a jest or cypress test to verify this?

I can see the lines of code are covered, but since you didn't have to change any tests when making this change, it means there isn't anything testing this specific case.

If testing this is hard because there aren't existing tests around this code I'm OK punting the test.

scytacki:
@dougmartin Something I just thought of is consistency. 

I think when the portal-report itself generates a link to open the standalone iframe view using `getStandaloneLinkUrl` it sets the studentId parameter to  `answer.platformUserId` which I think is the runKey for anonymous answers.

So in some cases these anonymous standalone iframe links will have a runKey and sometimes they will have a studentId which is a runKey.

So for consistency either:
- that code in the portalReport that generates those links for anonymous users should send the runKey and not the studentId. (this will make that code more complicated)
- the change in this PR should be reverted and code in the AP and LARA that is generating these links should be updated to send the answer.platformUserId as the studentId when generating the links.

It seems the second option is better, but I haven't looked at the places where the links are being generated.

scytacki:
I talked this over with Noah. And we realized that when the main portal report is launched it is already doing something similar to this PR. It looks for a runKey parameter and then looks for anonymous answers. 

So I think we should just leave this PR as it is. Noah is going to try to add a test of the changes before we merge it.

As we identified, this stuff is really confusing. In the case of the code in this PR, it is filtering answers by the `platformUserId`. In the other cases where runKey's are used it is filtering answers documents by their run_key here: https://github.com/concord-consortium/portal-report/blob/ffe9ef1ba38bd3ac0307d2978631e9a98f6a890b/js/actions/index.ts#L356 
Since the the runKey is stored in both places of the answer either approach works.

I tried checking the existing code to see what worked and what doesn't:
As we know without a studentId it fails like this link:
https://portal-report.concord.org/branch/master/index.html?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Factivities%2F21143&firebase-app=report-service-dev&iframeQuestionId=mw_interactive_211266&resourceUrl=https%3A%2F%2Fauthoring.staging.concord.org%2Factivities%2F21143&sourceKey=authoring.staging.concord.org&runKey=c69e25d1-5199-4cd7-8183-53c53adf7a9d

But without a runKey it also fails like this link:
https://portal-report.concord.org/branch/master/index.html?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Factivities%2F21143&firebase-app=report-service-dev&iframeQuestionId=mw_interactive_211266&resourceUrl=https%3A%2F%2Fauthoring.staging.concord.org%2Factivities%2F21143&sourceKey=authoring.staging.concord.org&studentId=c69e25d1-5199-4cd7-8183-53c53adf7a9d

And looking at what happens when looking at what happens with `getStandaloneLinkUrl`, the resulting URL has both a studentId and runKey parameter and that is why it works. This happens because that function is starting the params with `const params = queryString.parse(window.location.search);`. And when the report is opened anonymously it has a runKey param so this is then passed through. 

Anyhow I understand how it is working now, and the PR is good the way it is.
It would be nice to simplify all of this at some point though.

dougmartin:
I had to re-request the review to remove your changes requested status.  Adding a test for this would not be straightforward in the current setup.

scytacki:
Oh i missed the second part of the class comment. :+1:

scytacki (approved):
Works for me.

PR #372, opened 2022-01-13
firebase asked us to upgrade analytics

scytacki:


scytacki:
I don't know if this is useful anymore, but it has almost been 2 years, so it doesn't seem worth keeping around. So I'm closing it. 
cc @dougmartin

PR #374, opened 2022-01-26
Add a key property to `ReportItemIframe` for each question.

knowuh:
Without the key property the Iframe content is not reliably updating when question selection changes.

[#180647562]
https://www.pivotaltracker.com/n/projects/2441249


dougmartin (approved):
Looks good :+1: 

We verified it worked during pairing to fix the bug and verified it broke again when we removed it during testing.

PR #376, opened 2022-01-31
Fix checkbox position under response details [#181037903]

dougmartin:
This was broken when the flex direction was changed to support the report-item interactives.


knowuh (approved):
LGTM, thanks!

PR #377, opened 2022-02-16
Fix drag-and-drop interactive does not look correct in report/dashboard.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181154848

[#181154848]

Adds a URL param for standalone iframes that can be used by question interactives to determine whether they should or shouldn't be scaled to fit the confines of the report/dashboard. Also makes some CSS adjustments to help handle scaled interactives.

These changes go along with [these related changes in question-interactives](https://github.com/concord-consortium/question-interactives/pull/163) and [these changes in LARA](https://github.com/concord-consortium/lara/pull/909).

### DIFF
--- a/css/report/iframe-answer.less
+++ b/css/report/iframe-answer.less
@@ -5,6 +5,16 @@
     width: 100%;
     display: flex;
     flex-direction: column;
+
+    &.scaled {
+      height: 100%;
+      max-height: 100%;
+    }
+  }
+  &.scaled {
+    height: auto;
+    max-height: 300px;
+    overflow: hidden;
   }
 
   .iframe-answer-header {
--- a/cypress/integration/standalone-iframe.spec.js
+++ b/cypress/integration/standalone-iframe.spec.js
@@ -25,7 +25,7 @@ describe("Opening stand-alone iframe question with saved state", function() {
   it("should show an iframe with the base url", function() {
     getByCypressTag("standaloneIframe").should("be.visible");
     cy.get("iframe").should("exist")
-      .and('have.attr', 'src', 'https://models-resources.concord.org/table-interactive/index.html');
+      .and('have.attr', 'src', 'https://models-resources.concord.org/table-interactive/index.html?view=standalone');
   });
 });
 
@@ -40,7 +40,7 @@ describe("Opening stand-alone open response answer", function() {
         .should("be.visible")
         .should("have.text", "test answer 1");
     });
-  })
+  });
 
   describe("When specifying an anonymous user answer run via runKey", () => {
     beforeEach(() => {
@@ -52,7 +52,7 @@ describe("Opening stand-alone open response answer", function() {
         .should("be.visible")
         .should("have.text", "test answer 1");
     });
-  })
+  });
 });
 
 describe("Opening stand-alone multiple choice answer", function() {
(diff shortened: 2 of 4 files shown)


emcelroy:
@knowuh @scytacki I used a URL param to pass information about whether the interactive should scale itself, but I'm wondering if I should rather have done something similar to how lara-interactive-api provides context about whether the interactive is being shown in a report? It's not clear to me how that's done, though, so I opted for the URL param option for the scaling. (In case it's not clear, using just the `report` property on the question-interactives side isn't enough because `report` is always true when the interactive is displayed in a report, even when you click the link to open the interactive in its own tab.)

scytacki:
We haven't used URL params for the host to customize interactives in other cases. I'm not entirely sure why not, but my guess is that we have to pass so much configuration that it would make the URL very large.  For consistency sake I think we should stick with our approach of using postMessage for configuration.

So I think you should change this to using a property on the initInteractive message. 
Calling in it `view: "standalone"` seems fine to me. We can leave it undefined for other cases.

Alternatively, maybe you could just base this on the width of the interactive. If it is in in report mode and it is wider than X then show the prompt and don't scale. If that works it seems easier to implement, and it generally seems reasonable to me that the prompt is hidden and the interactive is scaled when there is limited space in the report.

If you want to add the new property to initInteractive you should change the code here: https://github.com/concord-consortium/lara/blob/03b23e0b661ed91de68023d0c573d1191083e6ce/lara-typescript/src/interactive-api-shared/types.ts#L91 and then need to release a new version of the interactive api client (for question-interactives) and interactive api host (for portal-report).

scytacki (changes requested):
See my previous comment for my suggestions. If you go with the width approach perhaps you don't have to make changes to the portal-report.

emcelroy:
@scytacki Thanks. I couldn't work out a way to use the width for determining when to scale the interactive. As I understand it, it's being displayed at the same size in both places, that's why it needs to be scaled down when its embedded in the report/dashboard UI.

I went the other route with the latest changes. I've updated [question-interactives](https://github.com/concord-consortium/question-interactives/pull/163) and [LARA](https://github.com/concord-consortium/lara/pull/909) accordingly.

Does that all look generally OK? I have the sense that lines 61-63 in iframe-standalone-app.js are kind of crude, but I'm not sure how else to add the view property.

scytacki (changes requested):
need to handle the case where the state is an object instead of a string

scytacki (approved):
Looks good to me.

css/report/iframe-answer.less
```diff
@@ -5,6 +5,16 @@
...
+
+    &.scaled {
+      height: 100%;
+      max-height: 100%;
+    }
+  }
+  &.scaled {
+    height: auto;
+    max-height: 300px;
+    overflow: hidden;
```
scytacki, line 17: @emcelroy can you describe how this is changing the behavior from before?
I normally skip over CSS changes, but here I wondering how it will effect the height when the interactive is requesting its own height. Some interactives might not look good if you give them the full height instead of what they ask for.

emcelroy: @scytacki These rules are applied to a wrapper for the iframe and its header. They're supposed to allow the interactive to use the full amount of space in the column (if it needs it) without overflowing it. 

image

Without the rules, the interactive can appear cropped. For example:

image

The rules don't affect the height of the actual interactive or it's iframe, AFAIK.

Does that seem reasonable? Or is it still possible that that could make things look bad? I did try this out with a variety of drag-and-drop interactives with different dimensions, aspect ratios, and aspect ratio settings. It seemed to work pretty well with them all, but it's possible I missed some combination that would've shown how this could be a problem.

PR #381, opened 2022-03-21
Fix drag-and-drop and side-by-side scaling in reports.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181154848

[#181154848]

Tweaks style rules to ensure proper scaling of interactive and that the entire scaled interactive is visible in the Response Details view in the dashboard. Without these changes, the drag-and-drop interactive was cut off a little at the bottom, and the side-by-side interactive wasn't being scaled down properly.

### DIFF
--- a/css/report/iframe-answer.less
+++ b/css/report/iframe-answer.less
@@ -13,7 +13,7 @@
   }
   &.scaled {
     height: auto;
-    max-height: 300px;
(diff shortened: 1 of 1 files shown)


knowuh (approved):
Thank you, looks good.

PR #385, opened 2022-05-11
Fix Playback Audio Answers in Class Dashboard

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181223927

[#181223927]

Open response questions with audio responses enabled should always show "No response" in the dashboard when a student has left neither an audio response nor a written response. If a student leaves only an audio response, the dashboard should show a button/link to play back the audio, and the text "No written response". If a student only leaves a written response, the dashboard should show just the text response (no playback button, no message about there being no audio).

Without these changes, open response questions with audio enabled would show "No written response" in the dashboard even if the student did not leave an audio response.


emcelroy:
@scytacki Do these changes seem like the right approach to fix the bug mentioned in the description? It seems that open response questions with audio enabled will still have an `answer` even when the student didn't provide a written or audio response. The answer's type is `interactive_state` and its `answer` property is essentially a state with an empty `interactiveState`. As I understand it, that's how it should be, but my fix here seems a little convoluted, which makes me doubt that.

PR #392, opened 2022-06-27
Fix view work/open in new tab links disappearing [#182562269]

dougmartin:
This moves the max-height from the enclosing container to the iframe content.  Without this the links were being hidden by the expanded iframe content.


emcelroy (approved):
Looks great 👍 Tested in a browser and worked as expected.

PR #393, opened 2022-07-11
Add support of "score" report item

pjanik:
[#182623252]

This PR is related to the new Interactive API: https://github.com/concord-consortium/lara/pull/982

It adds support of optional "score" report item. Report will ask interactive for "compactAnswer" items only when interactive provides report item metadata with `compactAnswerReportItemsAvailable=true`. This is done that way so we don't force older interactives like Sage or Geocode to render their report items 20-30 times (for each student) for no reason (they don't provide scores anyway). 

Once this approach is approved, I'll add more tests to increase the coverage.


package.json
```diff
@@ -64,7 +64,6 @@
   "devDependencies": {
     "@cypress/code-coverage": "^3.9.7",
     "@istanbuljs/nyc-config-typescript": "^1.0.1",
-    "@mermaid-js/mermaid-cli": "^8.8.4",
```
scytacki, line 67: Was mermaid causing problems? I believe it is needed to run `npm run doc:diagrams`

scytacki (changes requested):
Looks good. 

Only issue I found was the removal of mermaid.

package.json
pjanik: Hm, I think I had some problems installing it locally, but when I added it back, it worked. Maybe some minor version upgrade helped.

pjanik:
@scytacki, mermaid is added back, and I've also added a few tests that focus on the most important logic - when we should issue "getReportItemItem" iframe phone call and when we should skip it.

scytacki (approved):
looks good to me

PR #395, opened 2022-08-03
fix report crash with null

scytacki:
Some interactives save a null interactive state. This get serialized as "null" in the reportState. When that happened the portal-report was crashing. 

While fixing this I found that in the changes in v4.11.0  interactive answers without a `reportState` field were being ignored in most cases. I updated `answers.json` to add the `reportState` field.
To keep this from happening again I updated the test to check that the answer table matches.


cypress/integration/portal-dashboard/portal-dashboard-response-table.spec.js
```diff
@@ -25,46 +25,85 @@ context("Portal Dashboard Response Table",()=>{
...
+    }
+  }
+
   it('verify we display the correct student answer icons',()=>{
+    const OR = aType("Open Response Completed");
+    const MC = aType("Multiple Choice NonScored Completed");
+    const Mx = aType("Multiple Choice Scored Incorrect");
+    const Mv = aType("Multiple Choice Scored Correct");
+    const IQ = aType("Image Question with Open Response Completed");
+    const IA = aType("Interactive Completed");
```
emcelroy, line 71: It seems like the choice of `IA` for the variable name was intentional even though I would've expected `IC`. I'm not sure what the A indicates, but I'm probably just not familiar with whatever it is.

emcelroy (approved):
Looks good to me 👍 I left one non-blocking comment.

cypress/integration/portal-dashboard/portal-dashboard-response-table.spec.js
scytacki: All of them are "Complete", so it seemed weird to add a C to this one when I didn't add one other places. The C in MC was for Choice.  The A was for "active". But I agree it is weird. I think I've seen Piotr use IA for interactives before. But if you have another 2 letter suggestion for Interactive I'm open to it.

I've got one more portal-report change that I'll open a new PR so I can apply your suggestion there.

scytacki: Maybe `IN`?

emcelroy: Yeah, that seems reasonable. But reading your explanation and looking at the code again, I think `IA` makes sense, too. The only other alternatives I can think of are `ID` ("Interactive Done") or `IF` ("Interactive Finished") but those seem potentially confusing, and you're using "Completed" elsewhere. I think it's fine as is, or if you switch it to `IN`.

scytacki: OK I'll just leave it then, easier than changing it. 🤷

scytacki:
@tejal-shah you might find the approach here for testing the dashboard answer table of items useful in other places. 

The error reporting isn't too bad. At least in the local cypress runner, you can hover over the failing test and it will highlight the cell in the table and the test error will show what type of answer it expects.

PR #396, opened 2022-08-03
add interactive question for report item

scytacki:
This removes the report item urls from the other interactives.
This makes it possible to test the other interactives the normal way.
It also updates the version of the generic report item interactive. Old version of the report item interactive didn't handle the update API.
So this fixes a crash that happened when viewing the student work of many of the interactives in the demo data in A2.


emcelroy (approved):
Looks good 👍

PR #397, opened 2022-08-22
Launch AP Teacher Edition with auth-domain param

pjanik:
[#181821831]

This PR adds auth-domain URL param to AP Teache Editon links.
It's related to: https://github.com/concord-consortium/activity-player/pull/378

I've also added rollover text to this button, which is a separate story, but touches the same code.

### DIFF
--- a/js/api.ts
+++ b/js/api.ts
@@ -118,7 +118,7 @@ export const initializeAuthorization = () => {
   return false;
 };
 
-const getPortalBaseUrl = () => {
+export const getPortalBaseUrl = () => {
   const portalUrl = urlParam("class") || urlParam("offering");
   if (!portalUrl) {
     return null;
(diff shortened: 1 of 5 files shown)


scytacki (approved):
Looks good to me. 

The only thing that perhaps should be discussed is the `auth-domain` param name. Why that instead of `authDomain`? I'm guessing it is consistent with some other code or standard?

I'm approving it as is because I don't want to hold it up and I'm guessing you have a good reason.

pjanik:
@scytacki, the only reason why I used `auth-domain` was to be consistent with the Portal Report that is using the same approach and URL param. I wasn't sure if it's better to be consistent with other URL param names in AP, or with another app (Portal Report). But it felt to me like being consistent with other apps makes more sense. If we're moving towards camelCase for all the URL params and it's more important, I can change that easily.

PR #398, opened 2022-08-24
Cleanup log messages

pjanik:
https://www.pivotaltracker.com/story/show/181821731

I just went through most of the log events in Dashboard and tried to organize them a bit better. Now we should log user UI action and not necessarily what happens in the app internals - eg you shouldn't see 4 events that correspond to one user action anymore. Also, I tried to add a bit more context, usually using log event name - eg. "QuestionOverlayChangeStudent" and "ResponseDetailsChangeStudent" should make it clear where in the app teacher was switching between students.

Trudi seems to be happy with the results:
https://concord-consortium.slack.com/archives/CGTUE5W0Y/p1661267417789939


dougmartin (approved):
Looks good :+1: 

I'm wondering if we should let Trudi and Paul know about the changes - I think Paul might have some scripts that ingest these log events.

pjanik:
Trudi is aware and she's been playing with these updated messages already. I'll ask her to talk to Paul to update him. I think she has her own document with all the messages listed.

PR #399, opened 2022-08-29
Fix audio responses no longer displayed in reports

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/183101947

[#183101947]

`storageName` was incorrectly being set to "reportItemAnswersCompact" when `itemsType` was undefined. 

I think an additional wrinkle is that the Open Response library interactive on authoring.concord.org doesn't currently have a Report Item value set.

The spec test changes are a slight improvement, but would not have caught the issue with audio responses not showing up in reports. I'll work on improving the tests in that regard after requesting some advice. More comments to come.

### DIFF
--- a/js/reducers/report-reducer.ts
+++ b/js/reducers/report-reducer.ts
@@ -253,7 +253,7 @@ export default function report(state = new ReportState({}), action?: any) {
       return processReportItemRequests(state);
     case SET_REPORT_ITEM_ANSWER:
       const answer = getAnswer(state, action.questionId, action.reportItemAnswer.platformUserId);
-      const storageName = action.reportItemAnswer.itemsType === "fullAnswer" ? "reportItemAnswersFull" : "reportItemAnswersCompact";
+      const storageName = action.reportItemAnswer.itemsType === "compactAnswer" ? "reportItemAnswersCompact" : "reportItemAnswersFull";
       const currentReportItemAnswer = (answer && state.getIn([storageName, answer.get("id")])) || null;
       const reportItemAnswerChanged = JSON.stringify(currentReportItemAnswer) !== JSON.stringify(action.reportItemAnswer);
       if (answer && reportItemAnswerChanged) {
--- a/test/components/report/iframe-answer-report-item_spec.js
+++ b/test/components/report/iframe-answer-report-item_spec.js
@@ -3,11 +3,6 @@ import { shallow } from "enzyme";
 import { Map } from "immutable";
 import { IframeAnswerReportItem } from "../../../js/components/report/iframe-answer-report-item";
 
-jest.mock("../../../js/components/report/iframe-answer-report-item-attachment", () => ({
-  IframeAnswerReportItemAttachment: () => {
-    return <div>MockIframeAnswerReportItemAttachment</div>;
-  }
-}));
 const renderLinkMock = jest.fn();
 
 const answerText = "testing 123";
@@ -36,7 +31,7 @@ const htmlItem = {
   type: "html"
 };
 
-const attachmentAnswer = Map([
+const audioAttachmentAnswer = Map([
   ["answer", answerText],
   ["answerText", answerText],
   ["attachments", {
@@ -45,8 +40,8 @@ const attachmentAnswer = Map([
     }
   }]
 ]);
-const attachmentItem = {
-  name: "answerAttachment",
+const audioAttachmentItem = {
+  name: "test_attachment",
   type: "attachment"
 };
 
@@ -73,9 +68,9 @@ describe("<IframeAnswerReportItem />", () => {
     expect(wrapper.find("iframe")).toBeDefined();
     expect(wrapper.find("iframe").props().srcDoc).toEqual(answerHtml);
   });
-  it("should render an attachment component for an item of type attachment", () => {
-    const wrapper = shallow(<IframeAnswerReportItem answer={attachmentAnswer} item={attachmentItem} answerText={answerText} renderLink={renderLinkMock} />);
-    expect(wrapper.html()).toEqual(expect.stringContaining("MockIframeAnswerReportItemAttachment"));
+  it("should render a 'Play audio response' button for an audio attachment response", () => {
+    const wrapper = shallow(<IframeAnswerReportItem answer={audioAttachmentAnswer} item={audioAttachmentItem} answerText={answerText} renderLink={renderLinkMock} />);
+    expect(wrapper.html()).toEqual(expect.stringContaining("Play audio response"));
   });
   it("should call renderLinks for an item of type links", () => {
     shallow(<IframeAnswerReportItem answer={linksAnswer} item={linksItem} answerText={answerText} renderLink={renderLinkMock} />);


emcelroy:
@scytacki I'm not sure how to update the automated tests so they would've caught this problem. Would it be good to update test/components/report/iframe-answer_spec.js by adding a test for an answer with an audio attachment? Or would it be better to update the tests in cypress/integration/portal-dashboard? I had a go at each, but ran into problems with both. So I'm hoping to get some advice on which would be better before spending more time on it.

scytacki (commented):
This seems OK to me. But I'm not familiar enough with this code to know the ramifications of switching the logic. I tagged @pjanik so he could take a look.

I think maybe we made a separate audio open response library interactive with the report item url instead of making all open response items use it, but I can't remember for sure.

emcelroy:
With the latest commit, I tried again to get the automated tests to catch the kind of problem being fixed here. I'm still  hitting walls, though, I think. The latest changes probably shouldn't stay, but I'm leaving them for now to help illustrate some things.

Seems like changes to the tests in test/components/report/ won't work because of how things have to be mocked. Maybe there's a way to check for the problem in test/reducers/report-reducer_spec.js?

With the Cypress tests, it seems like one of the answer data files will have to reference an actual attachment. I haven't been able to make that work yet, but maybe it's a bad idea anyway? Seems like pointing to an actual file in S3 could be problematic.

js/reducers/report-reducer.ts
```diff
@@ -253,7 +253,7 @@ export default function report(state = new ReportState({}), action?: any) {
       return processReportItemRequests(state);
     case SET_REPORT_ITEM_ANSWER:
       const answer = getAnswer(state, action.questionId, action.reportItemAnswer.platformUserId);
-      const storageName = action.reportItemAnswer.itemsType === "fullAnswer" ? "reportItemAnswersFull" : "reportItemAnswersCompact";
+      const storageName = action.reportItemAnswer.itemsType === "compactAnswer" ? "reportItemAnswersCompact" : "reportItemAnswersFull";
```
pjanik, line 256: @emcelroy,

> In the report reducer, storageName was incorrectly being set to "reportItemAnswersCompact" when itemsType was undefined. As I understand it, storageName should be set to "reportItemAnswersFull" in that situation.

This change fixes the problem you described, but I'd say it's pretty hidden. Initially, I thought nothing has changed here. :) I can imagine it might be randomly changed by another developer not aware of this logic.

I'd vote to change it to a more explicit statement with `if` where we check for undefined value first. And/or some comment could be useful.

test/components/report/iframe-answer_spec.js
```diff
@@ -3,7 +3,18 @@ import { shallow } from "enzyme";
...
+      {
+        type: "attachment",
+        name: "test_attachment"
+      },
+      {
+        type: "answerText"
+      }
+    ];
+  }
+);
```
pjanik, line 17: `getReportItemAnswerMock` is used to send a postMessage request, it doesn't really return anything. So, this value here will be ignored by the IframeAnswer component. It could be an empty function too. I guess it'd be better not to cause any confusion.

IframeAnswer expects results to be available in the `reportItemAnswer` prop.

pjanik (approved):
> In the report reducer, storageName was incorrectly being set to "reportItemAnswersCompact" when itemsType was undefined. As I understand it, storageName should be set to "reportItemAnswersFull" in that situation.

That's correct. I'm surprised I didn't catch that, as I was testing OR question when I was developing ScoreBOT report item.

I've opened a small PR in OR that should also fix this problem on the OR side:
https://github.com/concord-consortium/question-interactives/pull/221
It's not necessary with your fix, but probably useful since OR question can be used as an example in the future.

Also, your fix is necessary for other old report-items to work (Sage, Geocoder, etc.).

> I think an additional wrinkle is that the Open Response library interactive on authoring.concord.org doesn't currently have a Report Item value set.

Yes, this should be definitely set.

> Seems like changes to the tests in test/components/report/ won't work because of how things have to be mocked. 

I think so. The bug was in the reducer logic that was picking the wrong storage / map. You won't be able to catch that in the IframeAnswer component specs if you already pass the reportItem there in the prop.

> Maybe there's a way to check for the problem in test/reducers/report-reducer_spec.js?

Yes, the problem was in the reducer logic, so that's the most logical place to add a unit test.

Some higher-level / integration tests could also work or be useful, but I don't remember now how they're organized here and how difficult it'd be.

> With the Cypress tests, it seems like one of the answer data files will have to reference an actual attachment. I haven't been able to make that work yet, but maybe it's a bad idea anyway? Seems like pointing to an actual file in S3 could be problematic.

I think it's a good idea. And I bet you don't even have to provide a real URL. It could be some bogus URL pointing to nowhere. That's probably enough for the report to render a valid link. Cypress won't really play it / validate if the sound is actually being played.

scytacki:
Yes I think a cypress test with an open response with a bogus audio URL seems like a great way to make sure this keeps working in the future. I'm not sure if we a way to mock the token-service or use the real thing, but that might be necessary to make this test work.

js/reducers/report-reducer.ts
emcelroy: @pjanik Are the latest changes to report-reducer.ts what you were thinking?

I couldn't think of a way to make it more concise, but maybe that's OK?

pjanik: My point is that I'd personally make it less concise. To handle `undefined` value in a very explicit way, so this logic is not broken accidentally later. For example:

```javascript
let storageName;
if (action.reportItemAnswer.itemsType === "compactAnswer") {
  storageName = "reportItemAnswersCompact";
} else if (action.reportItemAnswer.itemsType === "fullAnswer") {
  storageName = "reportItemAnswersFull";  
} else { 
  // Old interactives might not send back itemsType. Storage name should default to "reportItemAnswersFull".
  storageName = "reportItemAnswersFull"
}
```

It might be personal, so ofc it's up to you. I just thought that `? :` construct makes the handling of `undefined` value a bit hidden.

emcelroy: @pjanik Thanks. I had actually tried something very similar to that first, but was getting an error: `Type 'string' is not assignable to type 'keyof IReportState'.` Looks like if we cast `storageName` to `keyof IReportState` that error goes away. See latest changes. Does that seem OK to you?

emcelroy:
I updated the existing cypress/portal-dashboard/portal-dashboard-response-details.spec.js file to check for an answer with an audio attachment. It relies on changes to js/data/answers.json and js/data/sequence-structure.js. With the latter changes, a "Play audio response" button appears in the demo dashboard where expected. And the test will fail if we revert the changes made to js/reducers/report-reducer.ts.

@scytacki Is the placement of that check OK? Or should there be a totally separate set of tests specifically for audio attachments?

scytacki (approved):
This looks great to me. I think the placement of the check in the existing test is fine.

js/reducers/report-reducer.ts
scytacki: One note is that based on the code coverage, it seems like there are no tests of the `reportItemAnswersCompact`. I'm a little surprised by that. But I guess it means the cypress tests do not cover this case of a compact answer.

I don't think adding that test needs to be part of this PR.

PR #400, opened 2022-09-13
Update ImmutableJS v3 to v4, fix new type errors, handle breaking changes

pjanik:
This was the second part of my dev day work (after not so successful attempt to remove ImmutableJS completely: https://github.com/concord-consortium/portal-report/tree/rm-immutablejs). 

I've updated ImmutableJS from v3 to v4 as their changelog claimed much better typing. This triggered quite a lot of errors that I've tried to fix. Also, I've fixed some places where types were specified incorrectly (eg Map instead of List) - in most cases after double-checking them in the runtime.

I've made sure that unit tests and Cypress tests work, and I ran the report locally with test data and tried to click through most of the UI. But still, I think it'd be good to do a quite careful general smoke test /QA after this gets merged to ensure that nothing is broken.

### DIFF
--- a/js/components/portal-dashboard/activity-navigator.tsx
+++ b/js/components/portal-dashboard/activity-navigator.tsx
@@ -1,12 +1,12 @@
 import React from "react";
-import { Map } from "immutable";
+import { Map, List } from "immutable";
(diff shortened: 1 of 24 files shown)


scytacki (approved):
Looks good to me. Thanks for working on it.

PR #401, opened 2022-09-14
Bar graphs in reports / class dashboard

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182994863

[#182994863]

These changes address issues revealed when adding bar graph sparklines to the Sensor Interactive report item.

* `IframeAnswer`'s `renderLink` function needs to be bound.
* Wrap `IframeAnswer` links in a containing DIV so they're not treated as flex items which wrap to multiple lines.

Required by [this PR](https://github.com/concord-consortium/sensor-interactive/pull/118).


scytacki (approved):
looks good to me

PR #403, opened 2022-09-26
Bar graphs in reports / class dashboard

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182994863

[#182994863]

Additional changes to fix the following problems:

* "View Work" iframe toggle link text was not changing to "Hide" when clicked and iframe was loaded.
* Cypress test was failing because of missing `data-cy=answerText` element.

### DIFF
--- a/js/components/report/iframe-answer.tsx
+++ b/js/components/report/iframe-answer.tsx
@@ -41,7 +41,8 @@ export class IframeAnswer extends PureComponent<IProps, IState> {
   }
 
   toggleIframe() {
-    this.setState({iframeVisible: !this.state.iframeVisible});
+    const { iframeVisible } = this.state;
+    this.setState({iframeVisible: !iframeVisible});
   }
 
   getLinkURL(answer: string) {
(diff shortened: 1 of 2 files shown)


js/components/report/iframe-answer.tsx
```diff
@@ -41,7 +41,8 @@ export class IframeAnswer extends PureComponent<IProps, IState> {
   }
 
   toggleIframe() {
-    this.setState({iframeVisible: !this.state.iframeVisible});
+    const { iframeVisible } = this.state;
+    this.setState({iframeVisible: !iframeVisible});
```
dougmartin, line 45: This is fine but just a note you can also use the callback form of `setState`, it works well for toggles:

`this.setState(prev => ({iframeVisible: !prev}))`

dougmartin (approved):
Looks good :+1:

js/components/report/iframe-answer.tsx
emcelroy: Thanks 👍 In the latest commit, I made that change (with a slight tweak).

PR #404, opened 2022-09-28
Bar Graphs in Reports / Class Dashboard

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/183454493

[#183454493]

This change addresses an issue identified by QA. See the last point in Saravanan's last comment of 9/21. 

It makes it so only the actual answer is shown in the compare view. (The compare view sets `alwaysOpen` to `true`.)

Currently, graph answers will be shown like this in the compare view:
image

This change makes it look like so:
image

### DIFF
--- a/js/components/report/iframe-answer.tsx
+++ b/js/components/report/iframe-answer.tsx
@@ -202,7 +202,7 @@ export class IframeAnswer extends PureComponent<IProps, IState> {
             {maybeAnswerTextOrLinks}
           </div>
         )}
-        {reportItemAnswerItems.map((item, index) => (
+        {!alwaysOpen && reportItemAnswerItems.map((item, index) => (
           <IframeAnswerReportItem
             key={index}
             item={item}


emcelroy:
@dougmartin, @scytacki Does this change look OK? It fixes the described problem for graph answers, but I'm not sure if it could cause problems for other question types with a report item? If you don't know offhand, I can investigate that.

scytacki (changes requested):
I don't know for sure. But I think there are some interactives like open response which don't support all features in their iframe report view. For example audio playback. So this change would not work well for those types of interactives.  You might also want to look at Sage Modeler. I think it reports metrics via the reportItems which Dan might want to be visible in the compare view.

I think we have a problem where the report can't tell if a question would prefer to use its iframe or not. There is the `question.get("displayInIframe")` which comes from LARA's `display_in_iframe`. Currently that seems to be just to handle legacy interactives that used to have a report_url (old Sage and CODAP mainly). Its value is `!has_report_url`.  I originally added it with the intent that authors could control this setting. We'd have to add it LARA in the library interactive model. And then we'd have change the logic of how it is computed. We can probably get rid of the `!has_report_url` because the AP doesn't support the main interactives that use a report_url.  So then this could be set to false for open response and any other interactive that wants to rely solely on its answerText and reportItems.  Because we are now relying on this, we'd have to republish all of the LARA activities so this `display_in_iframe` property would be set to false for all of the activity structures in the report-service Firestore.

I don't know if this solution would help for the Sage modeler metric case though. In that case it might want to show both an iframe and the metrics.

I doubt it is worth doing this now. It should be documented, and explained to Leslie and Kiley. Screenshots of what happens currently with Open Response, Sage Modeler, and the bar graph, and what happens with the current solution of this PR would be good.  

Regardless of the solution, it might be worth renaming `alwaysOpen` to be `compareView` since that seems to be the only place that `alwaysOpen` is true.

dougmartin:
@emcelroy I think we can close this without merge as it was fixed in another PR and merged.  The history is in the PT story:

https://www.pivotaltracker.com/story/show/183454493

PR #405, opened 2022-10-17
Added injected style tag to html report item answers [#183557683]

dougmartin:
Without this style tag injected both the tall and wide html responses are displayed.


scytacki (approved):
looks good to me.

PR #407, opened 2022-10-24
Fix display of report item links in Compare View.

lublagg:


dougmartin (approved):
Looks good to me :+1: 

I think you should wait for @scytacki to also review since we paired to do this work.

js/components/report/iframe-answer.tsx
```diff
@@ -213,7 +214,7 @@ export class IframeAnswer extends PureComponent<IProps, IState> {
             {maybeAnswerTextOrLinks}
           </div>
         )}
-        {reportItemAnswerItems.map((item, index) => (
+        {!compareView && reportItemAnswerItems.map((item, index) => (
```
scytacki, line 217: Have you checked what happens with open response questions in this case?
As I wrote in the other PR. I think this will prevent open response questions from showing all of there features. It will also prevent something like SageModeler from showing its metrics in the compare view.

scytacki (changes requested):
I think this will suffer all of the same problems as the original PR.
https://github.com/concord-consortium/portal-report/pull/404/files

It just makes it more explicit that that this flag is for the compare view. But I think there are several interactives that should be showing their report items in the compare view.

scytacki:
@lublagg and @dougmartin I think the best incremental work to be done on this is to document the cases. Perhaps I'm wrong, but if I'm right then have this documentation will make it easier for devs and owners to understand how this isn't a simple problem.

So making screenshots of what happens with and without this PR for Open Response, Multiple Choice, Sage Modeler, and the bar graph.

dougmartin:
Closing - dead PR.

PR #408, opened 2022-10-26
[#18367077] Removed  code that renders null link.

lublagg:


js/components/report/iframe-answer.tsx
```diff
@@ -88,19 +63,18 @@ export class IframeAnswer extends PureComponent<IProps, IState> {
...
-    const linkUrl = this.getLinkURL(answer.get("answer"));
     const externalLinkIcon = <span className="pr-icon-external-link" />;
     if (question.get("displayInIframe")) {
       const toggleText = iframeVisible ? "Hide" : "View Work";
       const standaloneLinkUrl = this.getStandaloneLinkUrl(question, answer);
       return (
         <div>
-          <a onClick={this.toggleIframe} target="_blank" data-cy="toggleIframe">{toggleText}</a> | {" "}
-          <a href={standaloneLinkUrl} target="_blank" data-cy="standaloneIframe">Open in new tab {externalLinkIcon}</a>
+          {options?.hideViewInline ? <></> : <a onClick={this.toggleIframe} target="_blank" data-cy="toggleIframe">{toggleText}</a>}
```
scytacki, line 72: This used to have a `| {" "}` after it to separate it from the Open in new tab text.
It is weird that the hide* options were ignored before. Did you find some case were they were needed? 
I'd tend to just remove them unless you've got a case where they are needed.

scytacki (changes requested):
This will remove the `|` separator from the 2 links, and I'm worried the hide* options will actually be misconfigured so actually using them might hide things we don't really want to hide.

js/components/report/iframe-answer.tsx
lublagg: I have seen the values set in report-item -- haven't seen them set to true, they are set to false in the sensor-interactive report-item, but I thought it was weird that they weren't being used if we are setting them in report-items for interactives. But I can remove them from the code.

scytacki: Sorry I was wrong before. I figured these were some old options that weren't relevant anymore. After your comment I looked more closely and saw that they are used by the report-item stuff.  I looked back through the commit history. I see that these options were added along with the report-item changes, but were never handled. Probably they were forgotten.  So I think handling them is the the right thing to do.

But you'll have to come up with a solution for the `| {" "}`.  If only one link is shown you don't need to show the `|`.  You could push the links into an array and then using one of the options here to add the `|`: https://stackoverflow.com/questions/34034038/how-to-render-react-components-by-using-map-and-join
Or you could just add explicit logic (only render `|` if both hide options are false).

scytacki (changes requested):
Sorry, I flip-flopped on this after looking at your response. I left a comment in that original thread.

scytacki (approved):
Looks good. 👍

PR #409, opened 2022-10-27
[#183628618]: Fix iframe-answer excessive re-rendering.

lublagg:


lublagg:
Hi @scytacki, I found that moving dispatch action to componentDidMount prevents the near-infinite re-rendering that was happening previously and loads the correct data. However, I'm not sure if there was a reason why the dispatch action was being called specifically within render function itself.

js/components/report/iframe-answer.tsx
```diff
@@ -40,6 +40,16 @@ export class IframeAnswer extends PureComponent<IProps, IState> {
...
+  componentDidMount(): void {
+    const { answer, question } = this.props;
+    const hasReportItemUrl = !!question?.get("reportItemUrl");
+    if (hasReportItemUrl) {
+      setTimeout(() => {
+      this.props.getReportItemAnswer(question.get("id"), answer.getIn(["student", "id"]) as string, "fullAnswer");
+      }, 0);
+    }
+  }
+
```
scytacki, line 52: I'd guess this re-rendering was actually making it so that updated student work was being automatically shown to the teacher without the teacher having to navigate away and come back. If I'm right this fix will break that "feature". 

I don't think this PR should fix that issue, but can you try verifying this?  Open a report for a class where that you are also in as a student and see changes by the student show up in the report.  If this was working in master and this PR changes this behavior we should just raise that in Slack and see if we should put another story about that.

scytacki (changes requested):
This looks good. I'd just like to know if it changes the "real-time" behavior of the report items.

js/components/report/iframe-answer.tsx
lublagg: Hey Scott, that is accurate. I will document the issue and bring it up in Slack.

PR #410, opened 2022-10-31
[#183628618]: Fixed iframe-answer updating state.

lublagg:


scytacki (changes requested):
Can you tell me more about this change?

I see that it now creating a new object instead of modifying the current object in place.
Does this somehow fix the re-rendering issue?

lublagg:
Hi @scytacki, yes. Because we were directly modifying the object instead of a creating a new object, the value of "reportItemAnswerChanged" in the SET_REPORT_ITEM_ANSWER action inside the report reducer was always true, which caused the constant re-rendering.

Screen Shot 2022-11-02 at 10 31 16 AM

scytacki (approved):
👍

PR #419, opened 2023-10-26
feat: Update audio icon [PT-185083661]

dougmartin:
This changes the audio code to use the same microphone icon used when recording audio.


lublagg (approved):
👍

PR #425, opened 2024-02-22
Do not show interactive iframe links when report item is available

pjanik:
https://www.pivotaltracker.com/story/show/186970929

I believe `if (!hasReportItemUrl) {` is better than `if (reportItemAnswerItems.length === 0) {`, as the `reportItemAnswerItems` array is empty when the report item is still loading and before the parent receives an answer with items. `!hasReportItemUrl` ensures that links are never shown when the report item is available.


dougmartin (approved):
Looks good :+1:

dougmartin:
Merging for QA review since @pjanik is on vacation.

PR #426, opened 2024-02-26
fix: Updated rubric table border color [PT-187021195]

dougmartin:
Updated rubric table border color to match latest Zeplin spec.


eireland (approved):
👍 LGTM

PR #427, opened 2024-02-27
feat: Added Activity Score Settings [PT-187021230]

dougmartin:
Adds the contents of the Activity Score Settings dialog.

NOTE: this does not save the settings, that will be in a follow on PR.


lublagg (approved):
Looks great! 👍

PR #429, opened 2024-02-29
feat: Make Activity-level the default for Feedback Report [PT-187132915]

dougmartin:
Changes the default from question to activity level in the feedback report.


lublagg (approved):
Looks good! 👍

PR #431, opened 2024-03-01
feat: Save Activity Score Settings [PT-187021230]

dougmartin:
Saves the activity score settings in Firebase.

This also adds a disableRubric query parameter to use during development and testing.


lublagg (approved):
Looks good 👍

PR #433, opened 2024-03-01
fix: Lifted feedback level state up to app state [PT-187132915]

dougmartin:
This fixes the feedback level being reset to the initial state when the response details page is re-mounted by moving the feedback level up to the top app state.


lublagg (approved):
Looks good 👍

PR #434, opened 2024-03-01
feat: Implement No Score setting [PT-187021303]

dougmartin:
When No score is enabled, does not display the score field in each student's row of the dashboard or the Avg Score in the Feedback Info panel at the top


lublagg (approved):
Looks good 👍

PR #435, opened 2024-03-04
fix: Autoselect the question feedback when the response details view is set to question [PT-187132915]

dougmartin:
This ensures we don't have a disabled view shown when we switch to the feedback view.


lublagg (approved):
Looks good 👍

PR #439, opened 2024-03-05
feat: Display Avg Score in Feedback Info panel [PT-187021319]

dougmartin:
Displayed in Feedback Summary panel above the Rubric Summary Icon when either Manual Score or Score with Rubric is enabled as the scoring setting.  Shows calculated average score across the class out of total possible score.


lublagg (approved):
Looks good to me 👍

PR #440, opened 2024-03-05
feat: Flesh out score with rubric [PT-187021252]

dougmartin:
This displays the rubric score when the question has been fully scored.  This also only shows the rubric score value in the table if the score type is rubric.

Finally this adds back code that was lost in a merge to check for a max score value of 999.


lublagg (approved):
Looks good 👍

PR #441, opened 2024-03-05
fix: Allow empty values in score input [PT-187021265]

dougmartin:
Since the max value in the settings modal can now be blank code was added to disable the save button in that case.

This also fixes a couple of css issues around value inputs:

1. Set the border radius to 8px instead of 4.
2. Fixed modal styling around max value to push it to the left.


lublagg (approved):
Looks great! 👍

PR #444, opened 2024-03-06
fix: Average rubric score calculation [PT-187021319]

dougmartin:
When a score is removed from the rubric the value is set to 0 and in that case the scored question count should not be increased.


dougmartin:
Merging without review as it is a simple one-line change.

PR #445, opened 2024-03-07
feat: Add confirmation on max score setting [PT-187021265]

dougmartin:
This adds a confirmation dialog if the max score is set to be less than existing student scores.


lublagg (approved):
Looks good 👍

PR #446, opened 2024-03-07
NO MERGE: Add extra rubric criteria based on query param

dougmartin:
This allows us to see how large rubrics are handled without authoring them.

ONLY FOR QA/DEMO - DO NOT MERGE!


dougmartin:
Closing with merge as no longer needed to testing.

PR #453, opened 2024-03-10
feat: Display processed markup for score hint [PT-187173896]

dougmartin:
This uses the existing react tooltip library to render the processed markup.


lublagg (approved):
LGTM 👍

PR #454, opened 2024-03-11
fix: Logging of empty and duplicate score values [PT-187192076]

dougmartin:
This fixes two issues:

1. Empty score values were not being logged.  A new logEmptyLabel option was added to support this.
2. The score value was always being logged when blurred instead of only logging when the score changed on blur.


lublagg (approved):
Looks great 👍

PR #457, opened 2024-03-13
feat: Change when hasBeenReviewed flag is set [PT-187220438]

dougmartin:
Changes:

1. Changed hasBeenReviewed flag from set when feedback is saved to set when feedback loaded from Firebase.
2. Updated feedback icons to also show when score is set and changed hasRubricFeedback to ignore 0 scores.
3. Changed getStudentFeedbacks() selector to not filter on hasBeenReviewed flag.


lublagg (approved):
Looks great! 👍

PR #459, opened 2024-03-19
fix: Previous activity info showing in the Activity Level Feedback section [PT-187021319]

dougmartin:
Previous activity info should not be displayed in the Activity Level Feedback section. This section should only display information related to the selected activity.

Also fixes feedback given badges in these scenarios:

- When scoring with a rubric, feedback given badge should not appear on the student until all rows of the rubric have a score in them. At that point, the feedback given badge should appear and the awaiting feedback count should update.
- When No Score or Autoscore is selected, user can enter text feedback OR complete rubric to mark a student as Feedback Given. At which point, the awaiting feedback count should update. If Rubric score is selected, user must complete the rubric to mark a student as Feedback Given and update the awaiting feedback count. In this case, feedback text field can be left blank and student will still be marked feedback given (as long as rubric is complete).
- When Manual Score is selected, the user must enter a score in the score field to mark a student as Feedback Given and the awaiting feedback count should update. In this case, feedback text field can be left blank and student will still be marked feedback given (as long as score field is complete).


js/reducers/feedback-reducer.ts
```diff
@@ -61,19 +75,30 @@ export default function feedback(state = new FeedbackState({}), action: any) {
...
+        settings,
+        hasScoredQuestions
+      });
       return state.set("activityFeedbacks", fromJS(actFeedbacks) as Map<any, any>);
     default:
       return state;
   }
 }
 
+export const updateHashBeenReviewed = (params: {activityFeedbackValues: any; settings: any; hasScoredQuestions: any}) => {
```
lublagg, line 89: Wondering if you meant to call this updateHasBeenReviewed instead of updateHashBeenReviewed?

lublagg (approved):
Looks good 👍 Just had one question about a variable name but not a big deal at all.

js/reducers/feedback-reducer.ts
dougmartin: Good catch - I'll update it.  How about `canIHashRevieweeeed`?

dougmartin:
@lublagg Updated here: https://github.com/concord-consortium/portal-report/pull/459/commits/e5a2b207d09044a4d5fda3957621df141185fa0a

PR #460, opened 2024-03-20
fix: Pull manual scores from feedbacks and not higher level scores attribute [PT-187021319]

dougmartin:
The feedbacks.scores attribute is not updated in the selector when the question changes so the scores need to be pulled from the individual student feedback which does change when the question is changed.


lublagg (approved):
Looks good 👍

PR #462, opened 2024-03-22
Activity Score Automation

saravanan-udayakumar:
Activity Score Settings Automation Coverage

### DIFF
--- a/cypress/integration/portal-dashboard/portal-dashboard-activity-score.spec.js
+++ b/cypress/integration/portal-dashboard/portal-dashboard-activity-score.spec.js
@@ -1,0 +1,402 @@
+import ActivityScore from "../../support/elements/portal-report/activity-score-settings";
+
+context("Portal Dashboard Anonymous Mode",() =>{
+
+  const score = new ActivityScore();
+
+  beforeEach(() => {
+    cy.visit("/?portal-dashboard");
+    cy.get('[data-cy=navigation-select]').click();
+    cy.get('[data-cy="list-item-feedback-report"]').should('be.visible').click();
+  });
+
+    it('verify activity level feedback is displayed by default',()=>{
+      cy.get('[data-cy=activity-level-feedback-button]').invoke("attr", "class").should("contain", "active");
+    });
+    it('verify activity level feedback content',()=>{
+      cy.get('[data-cy=feedback-note-toggle-button]').invoke("attr", "title").should("contain", "Note on activity-level feedback");
+      cy.get('[data-cy=feedback-settings-toggle]').should("exist");
+      cy.get('[data-cy=feedback-settings-toggle]')
+      .should("contain", "Activity Score:")
+      .should("contain", "rubric");
+      cy.get("[class^='feedback-legend--feedbackBadgeLegend__rubric_score--']").should("exist");
+      cy.get("[class^='feedback-legend--feedbackBadgeLegend__rubric_score_avg--']")
+      .should("contain", "Avg. Score:")
+      .should("contain", "0 / 6");
+      cy.get("[class^='feedback-legend--feedbackBadgeLegend__rubric_summary']")
+      .should("contain", "Rubric Summary:")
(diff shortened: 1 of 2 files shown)


cypress/support/elements/portal-report/activity-score-settings.js
```diff
@@ -0,0 +1,79 @@
+import { getByCypressTag } from "../../../utils";
+
+class ActivityScore {
+
+  verifyActivityScoreSettingsOptionLabel(setting, label) {
+    const option = ["No score", "Manual", "Rubric", "MCQ Scored"];
```
johnTcrawford, line 6: I'm wondering whether this ought to be defined in the class rather than in each method; though it's only used 3 times, so it's debatable.

cypress/support/elements/portal-report/activity-score-settings.js
```diff
@@ -0,0 +1,79 @@
...
+  verifyScoreNotDisplayedInRubricSummaryDetailsDialogHeader() {
+    this.getRubricSummaryDetailsDialog().find('[class^=rubric-table--rubricScoreHeader--]').eq(0).should("not.contain", "3");
+    this.getRubricSummaryDetailsDialog().find('[class^=rubric-table--rubricScoreHeader--]').eq(1).should("not.contain", "2");
+    this.getRubricSummaryDetailsDialog().find('[class^=rubric-table--rubricScoreHeader--]').eq(2).should("not.contain", "1");
+  }
+  verifyScoreDisplayedInRubricSummaryDetailsDialogHeader() {
+    this.getRubricSummaryDetailsDialog().find('[class^=rubric-table--rubricScoreHeader--]').eq(0).should("contain", "3");
+    this.getRubricSummaryDetailsDialog().find('[class^=rubric-table--rubricScoreHeader--]').eq(1).should("contain", "2");
+    this.getRubricSummaryDetailsDialog().find('[class^=rubric-table--rubricScoreHeader--]').eq(2).should("contain", "1");
+  }
```
johnTcrawford, line 59: I'm wondering if the use of a method like this would make the above lines more legible:
  verifyScoreDisplayInRubricSummaryDetailsDialogHeader(index, containOrNot, string) {
    this.getRubricSummaryDetailsDialog().find('[class^=rubric-table--rubricScoreHeader--]').eq(index).should(containOrNot, string);
  }
Then each of the lines above would be something like this:
  verifyScoreDisplayInRubricSummaryDetailsDialogHeader(0, "not.contain", "3");
(You could also make containOrNot a boolean rather than a string, perhaps.)
It's a close call: my suggestion actually increases the number of lines of code, but makes most of those lines easier to read, I think.

cypress/integration/portal-dashboard/portal-dashboard-activity-score.spec.js
```diff
@@ -0,0 +1,402 @@
...
+  });
+
+    it('verify activity level feedback is displayed by default',()=>{
+      cy.get('[data-cy=activity-level-feedback-button]').invoke("attr", "class").should("contain", "active");
+    });
+    it('verify activity level feedback content',()=>{
+      cy.get('[data-cy=feedback-note-toggle-button]').invoke("attr", "title").should("contain", "Note on activity-level feedback");
+      cy.get('[data-cy=feedback-settings-toggle]').should("exist");
+      cy.get('[data-cy=feedback-settings-toggle]')
+      .should("contain", "Activity Score:")
```
johnTcrawford, line 20: I think that some of the values like '[data-cy=activity-level-feedback-button]' and '[data-cy=feedback-settings-toggle]' that are used 5 or 10 times should either be constants or should be put in methods to be called multiple times. The methods could be here in portal-dashboard-activity-score.spec.js, or perhaps in activity-score-settings.js instead.

johnTcrawford (approved):
Looks very good. I did make a few small suggestions, but you can take them or leave them - your decision. I approve either way.

PR #464, opened 2024-03-28
feat: Add secondary researcher header element [PT-187325112]

dougmartin:
This adds a secondary header element if the user is a researcher.  This new element displays under the main header and shows that the interface is in researcher mode and displays the class name and teacher names.

The teachers of the class were previously not storied in the state so most of the work of this commit is adding that to the state and drilling it down as a prop.

### DIFF
--- a/css/portal-dashboard/header.less
+++ b/css/portal-dashboard/header.less
@@ -269,3 +269,9 @@
   }
 }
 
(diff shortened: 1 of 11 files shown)


pjanik (approved):
Looks good 👍🏻

PR #465, opened 2024-03-28
feat: Audio badges in Progress Dashboard [PT-187279104]

dougmartin:
This adds a generic method to add badges to the progress dashboard and implements adding audio attachment badges to open response questions.


pjanik (approved):
Looks good 👍

PR #467, opened 2024-03-29
fix: Remove newly added teacher names to researcher header [PT-187325112]

dougmartin:
This removes the teachers names that were part of the initial new researcher sub-header as they are already displayed in the top right corner of the app.

Most of this PR is to remove the teachers from the state, which was added to show the teacher names in the header.


pjanik (approved):
Looks good 👍

PR #468, opened 2024-03-29
fix: Show researcher header outside progress dashboard view [PT-187325112]

dougmartin:
The response details and feedback views use fixed elements under the header and have a top value set which was covering up the optional researcher header.

This also fixes the bottom 3px white border (defined using a bottom-margin) disappearing when the user switched out of the progress dashboard.


pjanik (approved):
Looks good 👍

PR #469, opened 2024-04-01
fix: Change audio answer badge check [PT-187279104]

dougmartin:
Previously the audio answer badge check looked at the attachments and if there was at least one audio attachment the badge was shown.  This was found to be incorrect in QA as when the audio is deleted the attachment remains but the interactive state is updated to remove the pointer to the attachment.

This commit changes to look at the interactive state within the report state.

This also fixes a missing key issue in the audio badges display that causes a warning.


lublagg (approved):
Looks good!

PR #472, opened 2024-07-08
this prevents a crash when image answer is a string

scytacki:


pjanik (approved):
Looks good 👍 I guess it's safe to merge it.

pjanik:
I changed the PR status, and I'm going to merge it. We still have corrupted answer objects in Firestore, so I think this will definitely be an improvement.

PR #473, opened 2024-07-09
feat: Add glossary audio links demo page [PT-187885690]

dougmartin:
This uses demo data and will be expanded in another story to load and play real recorded glossary audio.


bacalj (approved):
Looks good to me - though I am not familiar with the actual app

css/glossary-audio/glossary-audio-app.less
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  .contents {
+    padding: 15px;
+
+    .info {
+      margin-bottom: 10px;
+    }
+  }
+}
```
bacalj, line 24: It's mad about no last line

dougmartin: Fixed in https://github.com/concord-consortium/portal-report/pull/473/commits/8b9484765c04fe47b8b24a049023649c4564cace

PR #480, opened 2024-09-05
fix: Student report logic for when to show rubric [PT-188024567]

dougmartin:
The old logic was preventing the text feedback or score from being shown if the rubric was hidden.


pjanik (approved):
Looks good 👍

PR #481, opened 2024-09-12
fix: Student feedback not showing rubric score [PT-188024567]

dougmartin:
The rubric score was not being shown when the teacher in the dashboard used the default score setting, which was to score using the rubric when the rubric was present.

This commit re-scores the rubric before it is shown to the student if there is no score type set or the score type is rubric and no score is set.

Further work needs to be done in the future to update the teacher dashboard to set the score by default.


pjanik (approved):
Looks good 👍

PR #483, opened 2024-10-25
feat: Updated sample rubric to v1.2 and added tag images and phrases [PT-188479846]

dougmartin:
Also added debug:rubricSummaryTableOverride query parameter to override the tag summary table rubric option.

### DIFF
--- a/js/core/rubric-migrations.js
+++ b/js/core/rubric-migrations.js
@@ -1,4 +1,10 @@
 import semver from "semver";
+import queryString from "query-string";
+
+function getSummaryTableOverride(currentValue) {
+  const rubricSummaryTableOverride = queryString.parse(window.location.search)['debug:rubricSummaryTableOverride'];
+  return rubricSummaryTableOverride ?? currentValue;
+}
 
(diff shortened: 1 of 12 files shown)


js/core/rubric-migrations.js
```diff
@@ -66,6 +80,7 @@ const migrations = [
     version: "1.2.0",
     migrations: [
       createCriteriaGroups,
+      fixUndefinedIconPhrase,
```
bgoldowsky, line 83: Since the version number will already be 1.2.0, it doesn't seem like this function will actually run when a rubric is fetched. Is that going to be a problem?

bgoldowsky (approved):
Looks good, I just had one question.

js/core/rubric-migrations.js
dougmartin: Good point - I'll fix that.

dougmartin: Actually I'll do it in the summary table display PR so I don't have to rebase and push the current follow on PT.

PR #491, opened 2025-01-16
feat: add title attribute for timestamp value (PT-188673526)

emcelroy:
[#188673526](https://www.pivotaltracker.com/story/show/188673526)


dougmartin (approved):
Looks good :+1:

PR #493, opened 2025-03-14
fix: OAuth breaking due to missing redirect protocol

dougmartin:


emcelroy (approved):
Looks good 👍

scytacki (approved):
Looks good to me. Probably we are passing the `auth-domain` url param without a scheme. But this approach seems like a fine work around.

PR #494, opened 2025-03-14
fix: Ensure all url parameters to urls have schemes

dougmartin:


scytacki (approved):
Looks good. Just need to fix the http:// handling later.

scytacki (approved):
looks good.

PR #496, opened 2025-04-30
feat: add deployment link to Jira (START-1)

emcelroy:
[START-1](https://concord-consortium.atlassian.net/browse/START-1)

Add `githubToken` and `deployRunUrl` to `.github/workflows/ci.yml`. This will allow the new version of `s3-deploy-action` to create a GitHub Deployment and set its `log_url` property, which will allow Jira to add a deployment link to associated Jira issues.

[START-1]: https://concord-consortium.atlassian.net/browse/START-1?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,8 @@ jobs:
           noPrefix: true
           awsAccessKeyId: ${{ secrets.AWS_ACCESS_KEY_ID }}
           awsSecretAccessKey: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          deployRunUrl: https://portal-report.concord.org/__deployPath__/
           # Parameters to GHActions have to be strings, so a regular yaml array cannot
           # be used. Instead the `|` turns the following lines into a string
           topBranches: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -76,6 +76,8 @@ jobs:
           noPrefix: true
           awsAccessKeyId: ${{ secrets.AWS_ACCESS_KEY_ID }}
           awsSecretAccessKey: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          deployRunUrl: https://portal-report.concord.org/__deployPath__/
```
emcelroy, line 80: In VSCode, there was a lint error for this `with` section: "Missing required input `prefix`". Even though `noPrefix` is present and set to true, `s3-deploy-action` still requires `prefix` ([see here](https://github.com/concord-consortium/s3-deploy-action/blob/main/action.yml#L14-L16)).

Seems like this is working without `prefix`, but should we add it? Or update `s3-deploy-action` so `prefix` isn't required?

scytacki: I would be good to update the s3-deploy-action so prefix isn't required.

scytacki, line 80: It'd probably be better to use this URL: 
```suggestion
          deployRunUrl: https://portal-report.concord.org/__deployPath__/?portal-dashboard
```
That way it will bring up the more commonly used dashboard instead of original report.

scytacki (approved):
I'm approving this, but it would be good to update the URL.

PR #497, opened 2025-05-09
feat: first activity is open by default (CLASSDASH-69)

emcelroy:
[CLASSDASH-69](https://concord-consortium.atlassian.net/browse/CLASSDASH-69)

If `currentActivity` is not set, we set it to be the first activity. This ensures the first activity is open by default.

Most of the files changed are Cypress test files that needed to be adjusted after the main change in `js/containers/portal-dashboard/portal-dashboard-app.tsx`.

[CLASSDASH-69]: https://concord-consortium.atlassian.net/browse/CLASSDASH-69?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/js/selectors/dashboard-selectors.js
+++ b/js/selectors/dashboard-selectors.js
@@ -98,7 +98,7 @@ export const getStudentAverageProgress = createSelector(
 export const getCurrentActivity = createSelector(
   [ getActivities, getCurrentActivityId ],
   (activities, currentActivityId) => {
-    return activities.find(activity => activity.get("id") === currentActivityId) || activities.first();
+    return activities.find(activity => activity.get("id") === currentActivityId);
   }
 );
 
--- a/js/selectors/dashboard-selectors.js
+++ b/js/selectors/dashboard-selectors.js
@@ -98,7 +98,7 @@ export const getStudentAverageProgress = createSelector(
 export const getCurrentActivity = createSelector(
   [ getActivities, getCurrentActivityId ],
   (activities, currentActivityId) => {
-    return activities.find(activity => activity.get("id") === currentActivityId);
+    return activities.find(activity => activity.get("id") === currentActivityId) || activities.first();
   }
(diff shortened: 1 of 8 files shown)


js/selectors/dashboard-selectors.js
```diff
@@ -98,7 +98,7 @@ export const getStudentAverageProgress = createSelector(
 export const getCurrentActivity = createSelector(
   [ getActivities, getCurrentActivityId ],
   (activities, currentActivityId) => {
-    return activities.find(activity => activity.get("id") === currentActivityId);
+    return activities.find(activity => activity.get("id") === currentActivityId) || activities.first();
```
dougmartin, line 101: The only problem with this approach is you can't toggle the activity to close it like you can currently on `master`.  I think that is useful to be able to see all the answers.

I think a better approach would be to set currentActivityId at load time to the first activity.  That way it can be toggled to undefined to close it like now.

dougmartin (changes requested):
Compare https://portal-report.concord.org/branch/first-activity-is-open-by-default/?portal-dashboard to https://portal-report.concord.org/branch/master/?portal-dashboard.  On master you can toggle the activity closed - in this new branch you can't.

I think setting currentActivityId on load time would be a better solution.

emcelroy:
> Compare https://portal-report.concord.org/branch/first-activity-is-open-by-default/?portal-dashboard to https://portal-report.concord.org/branch/master/?portal-dashboard. On master you can toggle the activity closed - in this new branch you can't.
> 
> I think setting currentActivityId on load time would be a better solution.

@dougmartin I see. Thanks for catching that. Does the new change look OK?

dougmartin (approved):
Looks good :+1:

PR #502, opened 2025-05-20
feat: Last Run column (CLASSDASH-67)

emcelroy:
[CLASSDASH-67](https://concord-consortium.atlassian.net/browse/CLASSDASH-67)

Adds a Last Run column to the Class Dashboard that shows when students last worked on an offering. The column can be hidden or shown by selecting a new "Hide Last Run column" option in the main hamburger menu.

[Demo on Branch URL](https://portal-report.concord.org/branch/last-run-column/index.html?portal-dashboard=true)

[CLASSDASH-67]: https://concord-consortium.atlassian.net/browse/CLASSDASH-67?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/css/portal-dashboard/level-viewer.less
+++ b/css/portal-dashboard/level-viewer.less
@@ -4,7 +4,7 @@
 @margin: 20px;
 
 .levelViewer {
-  min-width: calc(100vw - 348px); // 100% of the screen minus the width of the sidebar
+  min-width: calc(100vw - @left-panel-width);
   border-bottom:1.5px solid @cc-teal-light2;
   position: relative;
   height: @top-nav-level-viewer-height;
--- a/css/portal-dashboard/level-viewer.less
+++ b/css/portal-dashboard/level-viewer.less
@@ -4,7 +4,7 @@
 @margin: 20px;
 
 .levelViewer {
-  min-width: calc(100vw - 240px);
+  min-width: calc(100vw - 348px); // 100% of the screen minus the width of the sidebar
   border-bottom:1.5px solid @cc-teal-light2;
   position: relative;
   height: @top-nav-level-viewer-height;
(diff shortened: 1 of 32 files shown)


js/api.ts
```diff
@@ -240,6 +241,26 @@ export function fetchPortalDataAndAuthFirestore(): Promise<IPortalRawData> {
...
+    }
+
+    if (classData.students) {
+      classData.students.forEach((student: any) => {
+        if (student.user_id && lastRunMap[student.user_id] !== undefined) {
+          student.last_run = lastRunMap[student.user_id];
+        }
+      });
+    }
+
```
emcelroy, line 263: I'm particularly interested in thoughts on this. It seemed like the best option to me, but required altering the `data` files used for tests and the demo view.

I think we could alternatively keep the `last_run` information in the offering data and look up the values whenever we need it while we're iterating over the students. But I'm not sure that would be better or even as good as this.

css/portal-dashboard/level-viewer.less
```diff
@@ -4,7 +4,7 @@
 @margin: 20px;
 
 .levelViewer {
-  min-width: calc(100vw - 240px);
+  min-width: calc(100vw - 348px); // 100% of the screen minus the width of the sidebar
```
dougmartin, line 7: Could we create a LESS variable to hold the sidebar width?

js/api.ts
dougmartin: Seems good to optimize here like you have done when you get the portal data. We could also update the portal to return the `last_run` in the class data.

dougmartin (approved):
Looks great :+1:

PR #514, opened 2026-01-30
fix: Restore report item answers in question level feedback [CLASSDASH-110]

dougmartin:
Two bugs were introduced in v4.25.0 (commit 0ea2814) that caused blank answers in the question level feedback panel when using Firebase data:

1. The questionIdToAnswerId map in the reducer could only hold one answer per question, so only one student's report item answer was stored. Restored the getAnswer() lookup by questionId and platformUserId.

2. IframeAnswer never called updateReportItemAnswerItems() on initial mount, leaving reportItemAnswerItems as an empty array.

See specs/CLASSDASH-110-missing-question-level-feedback-answers.md for full analysis including answer history compatibility verification.


emcelroy (approved):
Looks good 👍

PR #517, opened 2026-05-20
feat: Add support email to the missing-structure error message [CLASSDASH-112]

dougmartin:
## Summary

Follow-up to #516. The "report isn't available yet" error shown to teachers when an activity's resource structure is missing now tells them who to contact.

Per a Slack request, the message ends with: *"If the problem continues, contact help@concord.org."* (previously "let Concord Consortium know if the problem continues").

## Notes

- `userMessage` is rendered as plain text in `DataFetchError`, so the address shows as readable text, not a clickable `mailto:` link.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


scytacki (approved):
looks fine

PR #518, opened 2026-05-23
feat: Support top-level multiple_answers on MC questions [CLASSDASH-114]

scytacki:
## Summary
- Read a new top-level `multipleAnswers` field on multiple-choice questions and fall back to `authoredState.multipleAnswers` so old report-structure documents keep working without a Firestore migration. (See [CLASSDASH-114](https://concord-consortium.atlassian.net/browse/CLASSDASH-114).)
- Encode multi-answer vs single-answer state into the choice icon's `data-cy` value (`multi-answer-choice-icon` / `single-answer-choice-icon`) so Cypress can verify which icon style rendered without inspecting SVG internals.
- Rename the dead `multi_answer` field in `activity-structure.json` to the new `multiple_answers` name (it was previously read by nothing).

## Background
LARA publishes the report structure as a single Firestore document and writes `authored_state` as a native Firestore object. Some valid JSON shapes aren't expressible as Firestore objects, which can break publishing (CLASSDASH-112 / LARA-212). LARA will start publishing `multiple_answers` at the top level of multiple-choice questions so the Class Dashboard doesn't need to crack open `authored_state` to find it. The portal-report's only read site for that field is [`multiple-choice-choices.tsx:101`](js/components/portal-dashboard/multiple-choice-choices.tsx#L101); this PR updates it to prefer the new field.

`??` (nullish coalescing) is used rather than `||` so an authored top-level `false` wins over a stale truthy value in `authoredState`.

## Test plan
- [x] Existing Cypress specs in `cypress/integration/portal-dashboard/*` and `cypress/integration/dashboard-multiple-choice.spec.js` all pass locally (130 tests, 129 passing, 1 pre-existing skip).
- [x] Two new Cypress assertions in `portal-dashboard-question-details.spec.js` cover both branches of the `??`:
  - top-level `multiple_answers: true` on `multiple_choice_22` (Q3) — verifies the new read path
  - `authored_state: { multipleAnswers: true }` on `multiple_choice_19` (Q2) — verifies the fallback path

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[CLASSDASH-114]: https://concord-consortium.atlassian.net/browse/CLASSDASH-114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/integration/portal-dashboard/portal-dashboard-question-details.spec.js
+++ b/cypress/integration/portal-dashboard/portal-dashboard-question-details.spec.js
@@ -163,6 +163,24 @@ context("Portal Dashboard Question Details Panel", () => {
       cy.get('[data-cy=overlay-class-response-area] [data-cy=multiple-choice-answers]')
         .should('have.text', 'Question doesn\'t have any choices');
     });
+    it('renders multi-answer icons when question has top-level multiple_answers (CLASSDASH-114)', () => {
+      // multiple_choice_22 (Q3) has top-level multiple_answers:true in the fixture.
+      cy.get('[data-cy="activity-question-button"]').eq(2).click();
+      cy.get('[data-cy=question-overlay]').should('contain', 'Question #3');
+      cy.get('[data-cy=overlay-class-response-area] [data-cy=multi-answer-choice-icon]')
+        .should('have.length.greaterThan', 0);
+      cy.get('[data-cy=overlay-class-response-area] [data-cy=single-answer-choice-icon]')
+        .should('not.exist');
+    });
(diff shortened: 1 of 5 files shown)


dougmartin (approved):
Approving. Logic is correct, both branches of the `??` are covered by Cypress, and CI is green. The `??` over `||` choice is the right call so an explicit top-level `false` wins over a stale truthy fallback.

Pushed a tiny follow-on commit (d900254) to drop the inner double-quotes around the new selectors so they match the unquoted `[data-cy=...]` style used elsewhere in the file. Pure style — feel free to squash on merge.

